PR #3, opened 2017-05-24
FlexBox React Component

ranmalr:
Added FlexBox React component with typescript and also added test cases for that.


src/components/FlexBox/FlexProps.tsx
```diff
@@ -0,0 +1,21 @@
+export const FlexAlign = {
```
harsimranb: Use `const enum` for all these.

PR #5, opened 2017-06-01
Video and Panel Component

krunalm:


src/components/Panel/Panel.tsx
```diff
@@ -0,0 +1,27 @@
...
+    theme?: any,
+    style?: React.CSSProperties,
+};
+
+export default class Panel extends React.PureComponent<Props, any> {
+    render() {
+        return (
+            <div className={this.props.theme ? this.props.theme.Panel : ''}>
+                {
+                    this.props.heading as string
```
harsimranb: It might be better to do: typeof this.props.heading === 'string'.

src/components/Video/Video.tsx
```diff
@@ -0,0 +1,91 @@
...
+                break;
+            default:
+                break;
+        }
+        return result;
+    }
+
+    render() {
+        return (
+            this.props.src.length === 1 ?
```
harsimranb: Is there anyway around this?  This `if` condition is prone to bugs, because most of the attributes are the same.

krunalm:
@harsimranb As discussed, have updated components for changes given. Have tested components, ran lint and tests on them and they worked fine.

PR #8, opened 2017-06-19
Theme support flex, video and panel components

krunalm:


src/components/Panel/Panel.scss
```diff
@@ -0,0 +1,11 @@
+.Panel {
```
harsimranb: make these lowercase

src/components/FlexBox/FlexBox.scss
```diff
@@ -0,0 +1,33 @@
+.Flex {
```
harsimranb: Make these classnames camelcase.

PR #35, opened 2017-07-26
30258-Link-UnitTest

DevanshiGandhi:
Link component unit-test.


src/components/Link/tests/Link.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+  it('is called when the Link is clicked', () => {
+      const spy = jest.fn();
+      shallow(<Link onClick={spy}>Test</Link>).simulate('click');
+      expect(spy).toHaveBeenCalled();
+  });
+});
+
+describe('external', () => {
+  it('should verify the external is false', () => {
+      const linkWrapper = mount(<Link external={false}/>);
```
mkulinski: `<Link external={false}/>` should have a space after } and before / `<Link external={false} />`

src/components/Link/tests/Link.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+      expect(linkWrapper.prop('external')).toBe(false);
+  });
+
+  it('should verify the external is true', () => {
+      const linkWrapper = mount(<Link external={true}/>);
+      expect(linkWrapper.prop('external')).toBe(true);
+  });
+
+  it('should verify the external type on the Link', () => {
+      const linkWrapper = mount(<Link external={true}/>).prop('external');
```
mkulinski: You don't need to pass a true value to a prop. If you just declare the prop with no value it will evaluate to true by default.

Example: 
`<Link external={true} />` === `<Link external />`

PR #48, opened 2017-08-02
Features/30329 create custom component card lists

HiirenP:
CardList component is now ready for review.


src/components/CardList/CardList.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import Item from './Item';
+
+export interface INodeItem {
```
mkulinski: interface properties on lines 4 - 13 should be separated by a semi-colon instead of a comma.
Ex:
`id: number;` not `id: number,`

src/components/CardList/CardList.tsx
```diff
@@ -0,0 +1,37 @@
...
+  selectedId?: number,
+}
+
+class CardList extends React.Component<Props, never> {
+
+  render() {
+
+    const nodes = this.props.data.map((person: INodeItem, index: number) => {
+      return (
+        <Item key={person.id || index}
```
mkulinski: <Item /> spaces off. Should look like this:
```
        <Item 
          key={person.id || index}
          node={person}
          children={person.people}
          selectedId={this.props.selectedId}
        />
```

src/components/CardList/CardList.tsx
```diff
@@ -0,0 +1,37 @@
...
+          node={person}
+          children={person.people}
+          selectedId={this.props.selectedId} />
+      );
+    });
+
+//indentation is not proper for List Component, So used <ul />, SO need to replace it with List.
```
mkulinski: When are we going to replace this with the <List /> component? 
Also, comments should start with a space.

src/components/CardList/Item.tsx
```diff
@@ -0,0 +1,87 @@
+import * as React from 'react';
+import Card from '../Card';
+import { INodeItem } from './CardList';
+
+export interface Props {
```
mkulinski: interface properties on lines 5 - 14 should be separated by a semi-colon instead of a comma.
Ex:
`id: number;` not `id: number,`

src/components/CardList/Item.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onToggle = () => {
+    const isToggled = !this.state.isToggled;
+    this.setState(() => ({
+      isToggled,
+    }));
+  }
+
+  render() {
+
+    let childnodes = null;
```
mkulinski: `childnodes` and `expandOrCollapse` should be in the component state instead of variables. Then they will be updated this the this.setState() method.

src/components/CardList/Item.tsx
```diff
@@ -0,0 +1,87 @@
...
+    let style: React.CSSProperties = {};
+    let expandOrCollapse: string;
+
+    const cardActions = [{ content: 'Select', onAction: this.onSelect }];
+    const children = this.props.node.people;
+    const collapseStyle: React.CSSProperties = { fontSize: '20px', color: 'green' };
+    const expandStyle: React.CSSProperties = { fontSize: '25px', color: 'red' };
+
+    if (this.state.isToggled) {
+      style = {
```
mkulinski: styles should be in a scss stylesheet and a className can be applied to component if there is a plus or minus. That way we can have a `cursor: pointer` and other styling applied throughout.

src/components/CardList/Item.tsx
```diff
@@ -0,0 +1,87 @@
...
+        );
+      });
+    }
+
+    //indentation is not proper for List Component, So used <ul />, SO need to replace it with List.
+    return (
+      <ul key={this.props.node.id}>
+        <span>
+          <Card title={this.props.node.name}  actions={cardActions}>
+            {childnodes ?  <a style={this.state.isToggled ? expandStyle : collapseStyle} onClick={this.onToggle}>{expandOrCollapse}</a> : null}
```
mkulinski: instead of using an `<a>` tag here, this should be using the `<Link />` component or broken out into another custom component.

PR #74, opened 2017-08-25
features/30729-Choice-Documentation

DimpleKundanani:
Documentation for Choice component is ready for code review.


docs/src/reducers/componentState/ChoiceState.ts
```diff
@@ -0,0 +1,49 @@
...
+import ChoiceExampleFirst from '../../examples/ChoiceExample/ChoiceExampleFirst';
+import ChoiceExampleSecond from '../../examples/ChoiceExample/ChoiceExampleSecond';
+
+const ChoiceExampleFirstCode = require('!raw-loader!../../examples/ChoiceExample/ChoiceExampleFirst') as string;
+const ChoiceExampleSecondCode = require('!raw-loader!../../examples/ChoiceExample/ChoiceExampleSecond') as string;
+
+const ChoiceState: IDocument = {
+  id: 'choice',
+  heading: 'Choice Component',
+  subheading: `Choice component is used to provide selection for RadioButton, Checkbox and Choicelist components.`,
```
mkulinski: why use ` ` instead of ' '?

DimpleKundanani: We have used Choice list component as an example for documentation.So it is there.

Thanks

PR #92, opened 2017-08-28
features/30724-Button-Documentation

DevanshiGandhi:
Documentation for Button component is ready for code review.


docs/src/examples/ButtonExample/ButtonExampleFirst.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+class ButtonExampleFirst extends React.Component<IProps> {
+  constructor(props: IProps) {
+    super(props);
+    this.state = {
+    };
+  }
+
+  valueUpdater(field: any) {
+    return (value: any) => this.setState({ [field]: value });
```
mkulinski: There is no state for this function call to alter.

docs/src/examples/ButtonExample/ButtonExampleFirst.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+  render() {
+    return (
+            <div className={styles.example}>
```
mkulinski: should only be two spaces of indentation after return.

docs/src/examples/ButtonExample/ButtonExampleFirst.tsx
```diff
@@ -0,0 +1,70 @@
...
+export interface IProps{
+}
+
+export interface IState {
+}
+
+class ButtonExampleFirst extends React.Component<IProps> {
+  constructor(props: IProps) {
+    super(props);
+    this.state = {
```
mkulinski: Since you're not using state at all, this can be removed from the constructor.

PR #118, opened 2017-08-29
Features/datepicker

kimhjona:


kimhjona:
hey @mkulinski ,
there are 72 files (which is a lot), but most of them are an exact clone of the airbnb datepicker. I already removed the testing files since we don't actually use them and can submit a separate PR for them.

kimhjona:
Also, I'll delete the other branch once this one goes through (just in case)

example/src/SingleDatePickerWrapper.jsx
```diff
@@ -0,0 +1,145 @@
+import React from 'react';
```
mkulinski: Why is this file here?

example/src/DateRangePickerWrapper.jsx
```diff
@@ -0,0 +1,144 @@
+import React from 'react';
```
mkulinski: Why is this file here?

package.json
```diff
@@ -132,6 +133,24 @@
     "velocity-animate": "^1.5.0",
     "redux": "3.6.0",
     "redux-immutable-state-invariant": "^2.0.0",
-    "redux-thunk": "2.2.0"
+    "redux-thunk": "2.2.0",
+    "@types/moment": "^2.13.0",
```
mkulinski: Are all of these needed?

src/components/DatePicker/components/OutsideClickHandler.jsx
```diff
@@ -0,0 +1,61 @@
...
+      { capture: true },
+    );
+  }
+
+  componentWillUnmount() {
+    if (this.clickHandle) removeEventListener(this.clickHandle);
+  }
+
+  onOutsideClick(e) {
+    // console.log('e:',e);
```
mkulinski: remove un-needed comments

src/components/DatePicker/index.ts
```diff
@@ -0,0 +1,53 @@
+// import undefined from '../../../lib/@types/src/components/DatePicker/utils/toMomentObject';
```
mkulinski: removed comments.

example/src/DateRangePickerWrapper.jsx
kimhjona: it's similar to the ReactDataGrid file - you need these two folders as a component to have a working calendar. I guess the reason it's like this is b/c I inherited it like this. 

We can talk about it in an hour or so

package.json
kimhjona: def not

PR #122, opened 2017-09-05
#30759 added validated form component in documentation

tejashrigajjar:


docs/src/examples/ValidatedFormExample/ValidatedFormExample.tsx
```diff
@@ -0,0 +1,78 @@
...
+              ]}
+            />
+            <ValidatedTextField
+              multiline
+              id="appDescription"
+              required={true}
+              name="App Description"
+              value={this.state.appDescription}
+              label="App Description"
+              placeholder=""
```
mkulinski: No need to put this here if it's going to be a blank string. Otherwise give it a value.

tejashrigajjar: I removed blank string props placeholder.

PR #125, opened 2017-09-07
Features/30728 chip component documentation

SandipMehta:


docs/src/examples/ChipExample/ChipExampleFirst.tsx
```diff
@@ -0,0 +1,77 @@
+import * as React from 'react';
+import { Chip } from '../../../../src/components/';
+import * as styles from '../../styles/components-page.scss';
+
+
+export interface IProps{
```
kimhjona: did you have a specific reason why have an empty props and state interface? you can just do something like React.Component<{}, {}> for line 17.

SandipMehta: There is not specific reason but we are keeping them same as our sample document example. Let me know in case we need to remove it so we will remove it from every component when it is not required. Thanks.

kimhjona:
hi sandip, thanks for your comments. I think it's fine that we have them. Can you merge with master? I will approve it once you do it.

SandipMehta:
I have resolved conflicts. Please approve it so I can merge it.

kimhjona:
looks good!

PR #144, opened 2017-09-26
Features/31802 text field counter length support

kimhjona:


example/src/App.tsx
```diff
@@ -373,13 +383,14 @@ class App extends React.Component<{}, State> {
...
         <Heading>Connected Text Field</Heading>
+        <Loading />
         <TextField
           label="Connected Text Field"
           type="number"
           placeholder=""
           value={this.state.appTextCounter}
           helpText="Helper Text"
-          enableTextCouter={true}
+          /* enableTextCouter={true} */
```
mkulinski: comment?

src/components/TextField/TextField.scss
```diff
@@ -7,6 +7,10 @@ $stacking-order: (
   backdrop: 10,
 );
 
+.red {
+  color: red;
```
mkulinski: is there a specific red we should make this? (Todd question)

kimhjona: screen shot 2017-09-27 at 1 08 22 pm
todd asked for red, but my guess is that the exact shade would matter too much in this case

mkulinski:
Please resolve merge conflicts and push changes

PR #154, opened 2017-10-11
Features/31963 picker component accessibility

kimhjona:


example/src/PickerAutoSuggestExample.js
```diff
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { PeoplePickerSearchType } from './PickerEnum';
 import { PeoplePickerSource } from './PickerSource';
```
mkulinski: Remove PickerSource and PickerEnum from component

example/src/PickerAutoSuggestExample.js
```diff
@@ -51,71 +54,119 @@ class PickerAutoSuggestExample extends React.Component<{}, {}> {
...
-            this.setState({
-              chipListState: this.state.chipListState,
-            });
-          }
-        } else if (this.state.chipListState.length && !this.state.value.length && !(e.keyCode === 38) && !(e.keyCode === 40)) {
-          if (this.state.chipListState[this.state.chipListState.length - 1]['markedForDelete']) {
-            this.state.chipListState[this.state.chipListState.length - 1]['markedForDelete'] = false;
-            this.setState({ chipListState: this.state.chipListState });
-          }
+      onfocus: (e:any) => {
```
mkulinski: Move methods needed for the component to work internally and give the users limited or optional access.

src/components/Chip/Chip.tsx
```diff
@@ -34,48 +36,47 @@ class Chip extends React.PureComponent<Props, any> {
...
-          theme.Chip,
-          transparent && theme.transparent
-      );
-    }
+    const deleteInstruct = this.props.removable ? 'Press delete to remove this chip' : '';
 
     const chipContents = [(
       image
       ?
-      <img className={theme.Image} src={image.url} alt={image.alt} key="1" />
```
harsimranb: @weegrom Do you know why the alt tag was removed?

src/components/Chip/Chip.tsx
```diff
@@ -34,48 +36,47 @@ class Chip extends React.PureComponent<Props, any> {
...
-        theme.markedForDelete,
-        transparent && theme.transparent
-      );
-    } else {
-      className = classNames(
-          theme.Chip,
-          transparent && theme.transparent
-      );
-    }
+    const deleteInstruct = this.props.removable ? 'Press delete to remove this chip' : '';
```
harsimranb: @weegrom Do you have any opinion on the wording of this text. We have this hidden in text below, probabaly for the readers.  Is team India going to test this kind of functionality? @krunalm

src/components/Chip/Chip.tsx
```diff
@@ -34,48 +36,47 @@ class Chip extends React.PureComponent<Props, any> {
...
     ];
-
     return (
-      <div className={className}>
+      <div
+        className={className}
+        tabIndex={this.props.tabIndex === 0 ? this.props.tabIndex : -1}
+        onFocus={this.props.autoSuggestMethods ? this.props.autoSuggestMethods.onfocus : ''}
+        onBlur={this.props.autoSuggestMethods ? this.props.autoSuggestMethods.onFocusOut : ''}
+        ref={this.props.autoSuggestMethods ? this.props.autoSuggestMethods.storeFocus : ''}
```
harsimranb: the ref is code smell, we should review and potentially reimplement this.

src/components/Chip/Chip.tsx
```diff
@@ -34,48 +36,47 @@ class Chip extends React.PureComponent<Props, any> {
...
       </span>,
     ];
-
     return (
-      <div className={className}>
+      <div
+        className={className}
+        tabIndex={this.props.tabIndex === 0 ? this.props.tabIndex : -1}
+        onFocus={this.props.autoSuggestMethods ? this.props.autoSuggestMethods.onfocus : ''}
+        onBlur={this.props.autoSuggestMethods ? this.props.autoSuggestMethods.onFocusOut : ''}
```
harsimranb: It would be cleaner if we create separate functions for onFucus and onBlur and check and class autoSuggestMethods there, instead of needing this logic here.

src/components/Chip/Chip.tsx
```diff
@@ -34,48 +36,47 @@ class Chip extends React.PureComponent<Props, any> {
...
+        ref={this.props.autoSuggestMethods ? this.props.autoSuggestMethods.storeFocus : ''}
+        role="option"
+      >
         {
           clickable
             ?
-            <a onClick={this.props.onClick} aria-disabled={false} tabIndex={0}>
-                {chipContents}
+            <a onClick={this.props.onClick ? this.props.onClick : this.props.autoSuggestMethods.onClick} aria-disabled={false} role={'alert'} tabIndex={!this.props.tabIndex ? 0 : -1}>
```
harsimranb: This onClick implementation logic is also not clean. I would suggest to have an onClick function in this component, which then calls the other logic.

src/components/Picker/AutoSuggestText.tsx
```diff
@@ -1,61 +1,43 @@
...
 import { TEXT_FIELD } from '../ThemeIdentifiers';
-
 import * as baseTheme from './TextField.scss';
-
 export interface Props {
-  itemsList?: object[];
   theme?: any;
   placeholder?: string;
   autoSuggestMethods: any;
   stateProps: any;
```
harsimranb: stateProps and autoSuggestMethods types need to be defined.  `:any` is bad, and I am getting code smell with "state" props here as well.  In general, it's better practice to have flat hierarchy in your props, vs. a nested structure. THat leaves everyone guessing, or browsing around the definition to figure out that hierarchy.

src/components/Picker/Card.tsx
```diff
@@ -19,13 +19,13 @@ class Card extends React.Component<Props, {}> {
     return (
       <div>
         <div className={ cardBackground }>
-          <span><img className={ style.avatarImage + ' ' + style.cardElem } src={ this.props.image } alt="not found" /></span>
```
harsimranb: @weegrom Hardcoded text like "not found" here and the one I mentioned above is not what we want. I think we should make this props so that we can do localization in the future, that's what we had discussed earlier. What do you think?

src/components/Picker/Picker.tsx
```diff
@@ -40,7 +39,6 @@ export interface Props {
   theme?: any;
   autoSuggest?: boolean;
   autoSuggestMethods?: object[];
```
harsimranb: What is this `type`?!

src/components/Picker/Picker.tsx
```diff
@@ -40,7 +39,6 @@ export interface Props {
...
-  itemsList?: object[];
   stateProps?: {chipListState: any, suggestions: any, inputProps: object, value?: any};
```
harsimranb: These types are sadly defined, I don't like passing state around as being done so many times in components above.

src/components/Picker/TextField.tsx
```diff
@@ -24,7 +24,6 @@ export interface Props {
   theme?: any;
   autoSuggest?: boolean;
   autoSuggestMethods?: object[];
-  itemsList?: object[];
   stateProps?: {chipListState: any, suggestions: any, inputProps: object, value?: any};
```
harsimranb: Same problemw with state props above

src/components/Picker/TextField.tsx
```diff
@@ -79,15 +78,13 @@ class TextField extends React.PureComponent<Props, State> {
     return (
       <Labelled
         label={label}
-        id={'textfield!!!'}
+        id={'textfield!!!' + label}
```
harsimranb: Why such a strange id?

harsimranb:
@weegrom @krunalm It's mind-boggling a task for "component accessibility" ends up with architectural changes in component implementation.

src/components/Picker/Picker.tsx
```diff
@@ -34,20 +33,17 @@ export interface Props {
...
   moreInfoComponentShowOn?: DisplayMoreInfo;
   style?: React.CSSProperties;
   theme?: any;
   autoSuggest?: boolean;
   autoSuggestMethods?: object[];
-  itemsList?: object[];
   stateProps?: {chipListState: any, suggestions: any, inputProps: object, value?: any};
   searchBehavior?(): void;
   onSelect?(item: any): void;
   onRemove?(item: any): void;
```
harsimranb: For item everywhere, instead of `any` for item type, we should look into using generics.  But, this is more a nice to have, if its too much work, its not worth it.

src/components/Chip/Chip.tsx
weegrom: @harsimranb It looks like this text was borrowed from Angular, so I'm okay with the wording.

@harsimranb and @krunalm  Yes, we do want the team testing accessibility.

src/components/Chip/Chip.tsx
weegrom: @harsimranb I'm just as confused as you as to why this line was removed.

src/components/Picker/Card.tsx
weegrom: @harsimranb  I 100% agree

example/src/PickerAutoSuggestExample.js
HiirenP: @harsimranb 
Do we need these changes? I guess PickerAutoSuggestExample is not related to Picker component at all. Because without having use of PickerAutoSuggestExample, we can use all features of Picker component. any thoughts on this?

src/components/Picker/TextField.tsx
HiirenP: @harsimranb  Maybe for a purpose of uniquely identify Text fields, when we have more than one on a page.

src/components/Picker/AutoSuggestText.tsx
HiirenP: @harsimranb Agree! but some of the methods in stateprops conflicts with other props and I receive a couple of warnings when I tried to change the method names. so as of now I implemented the interface and defined stateprops so functionality wise it works perfectly without changing too much of code.

example/src/PickerAutoSuggestExample.js
harsimranb: @HiirenP That's an interesting question.  I do think that this example has the people picker implementation we will be using.  You can check with @weegrom as well.

src/components/Picker/TextField.tsx
harsimranb: @HiirenP I guess it's fine.  I'm not a fan of "!!!".  Also, we could perhaps using `textfield!!!${label}` syntax, but this is not a big deal.

src/components/Picker/Picker.tsx
```diff
@@ -77,12 +92,179 @@ class Picker extends React.Component<Props, State> {
       searchItems: [],
       selectedItems: [],
       moreInfo: false,
+      value: '',
+      input: () => { },
+      suggestions: [],
+      chipListState: [],
+      focusArr: [],
+      itemsList:
```
krunalm: Do not add dummy data to component. Instead use data from source props directly. You may remove itemsList also.

src/components/Picker/Picker.tsx
```diff
@@ -77,12 +92,179 @@ class Picker extends React.Component<Props, State> {
...
+          if (number === chipListState.length) focused = number - 1;
+          else if (number > 0) focused = number;
+          else focused = 0;
+          if (focusArr.length) focusArr[focused].focus();
+          else this.state.input.focus();
+          this.setState({ focused });
+        }
+      },
+      onClick: (e: React.FormEvent<Element>) => {
+        console.log('clicked!');
```
krunalm: No need for console.log. Do we need onClick ?

src/components/Picker/Picker.tsx
```diff
@@ -15,15 +16,24 @@ export interface IPickerInfo {
 }
 
 export interface IStateProps {
```
krunalm: IStateProps is not used? Do we need it?

src/components/Picker/Picker.tsx
```diff
@@ -77,12 +92,179 @@ class Picker extends React.Component<Props, State> {
...
+      },
+    };
+    const { value, suggestions, chipListState } = this.state;
+    const inputProps = {
+      value,
+      onChange: autoSuggestMethods.onChange,
+      onFocus: autoSuggestMethods.onInputFocus,
+      'aria-label': 'Picker input',
+    };
+    const stateProps: any = { value, suggestions, chipListState, inputProps };
```
krunalm: Try to use IStateProps over here and match types for the FormEvent.

example/src/PickerAutoSuggestExample.js
```diff
@@ -4,186 +4,14 @@ import AutoSuggestText from '../../src/components/Picker/AutoSuggestText';
 import Card from '../../src/components/Picker/Card';
```
krunalm: Remove this references as it is not used.

krunalm: PickerAutoSuggestExample.js should be converted to typescript as well.

src/components/Chip/Chip.tsx
```diff
@@ -18,85 +17,54 @@ export interface Props {
...
+  onClick?(event: any): void;
   handleMoreInfo?(): void;
-  key?: number;
-  tabIndex?: number;
-  autoSuggestMethods?: IAutoSuggestMethods;
-  onfocus?(event: Event): void;
-  id?: number;
+  key?: any;
+  markedForDelete?: boolean;
+  children?: string;
```
krunalm: Sync up chip component with Documentation available in docs app. Both should match and if required update docs

src/components/Chip/Chip.tsx
```diff
@@ -18,85 +17,54 @@ export interface Props {
   moreInfoComponentShowOn?: DisplayMoreInfo;
   style?: React.CSSProperties;
   theme?: any;
-  onRemove?(event: React.FormEvent<HTMLElement>): void;
-  onClick?(event: React.FormEvent<HTMLElement>): void;
+  onRemove?(event: any): void;
```
krunalm: Chip onremove and onclick - event type changed to any - revert it and check

src/components/Chip/Chip.tsx
```diff
@@ -18,85 +17,54 @@ export interface Props {
...
+  onClick?(event: any): void;
   handleMoreInfo?(): void;
-  key?: number;
-  tabIndex?: number;
-  autoSuggestMethods?: IAutoSuggestMethods;
-  onfocus?(event: Event): void;
-  id?: number;
+  key?: any;
```
krunalm: Remove key prop as it is not used.

src/components/Picker/Picker.tsx
```diff
@@ -246,6 +195,26 @@ class Picker extends React.Component<Props, State> {
...
+
+      storeFocus: (e: HTMLElement) => {
+        if (!this.state.focusArr.includes(e) && e !== null) {
+          const focusArr = this.state.focusArr.length ? this.state.focusArr.concat([e]) : [e];
+          this.setState({ focusArr });
+        }
+      },
+
+      onClick: (e: React.FormEvent<Element>) => {
+        // TODO can write a code accoring to requirements
```
krunalm: Picker click event remove it as discussed and make sure it works

src/components/Picker/tests/UnthemedPicker.test.tsx
```diff
@@ -7,40 +7,60 @@ const theme = {
...
       const subject = mount(<UnthemedPicker
-        source={searchData}
+        source={Data}
+        autoSuggest
         chipComponent={Chip}
         searchResultComponent={Chip}
-        searchBehavior={spySearch}
-        onSelect={spyClick}
+        searchBehavior={spySearch()}
+        onSelect={spyClick()}
```
krunalm: Spy function should be used like this  `searchBehavior={spySearch} ` only. Remove '()' call from it and see actual unit test failing issue. 

This is applicable to both Picker and Chip Unit tests. Do it for both components tests.

src/components/Chip/Chip.tsx
```diff
@@ -49,22 +51,28 @@ class Chip extends React.PureComponent<Props, any> {
...
       : chipContents;
     const isRemovable = removable ?
-      <button type="button" className={theme.Remove} aria-label={'Remove ' + children} onClick={onRemove}>
+      <a className={theme.Remove} aria-label={'Remove ' + children} onClick={onRemove} tabIndex={-1}>
         <span aria-hidden="true">×</span>
-      </button>
+      </a>
       : '';
-    const returnedValue = clickable || removable ? React.createElement('span', { className }, isClickable, isRemovable) : React.createElement('button', { className }, isClickable, isRemovable);
+    const returnedValue = clickable || removable ? React.createElement('button', { className }, isClickable, isRemovable) : React.createElement('button', { className }, isClickable, isRemovable);
```
krunalm: Both conditions are now same. So, remove returned value variable and use single button react element creation.

src/components/Chip/Chip.tsx
HiirenP: resolved

src/components/Chip/Chip.tsx
HiirenP: resolved

src/components/Picker/tests/UnthemedPicker.test.tsx
HiirenP: resolved

src/components/Picker/Picker.tsx
HiirenP: resolved

src/components/Chip/Chip.tsx
HiirenP: resolved

src/components/Chip/Chip.tsx
HiirenP: resolved

PR #155, opened 2017-10-27
Features/32663 picker bug changes

HiirenP:
32633_picker Bug_changes_with docs


krunalm:
As discussed with @HiirenP, this code is merged with PR #154 31963 
Hence it is closed.

PR #157, opened 2017-10-27
solved bug and added resizable property

shyamk9:


weegrom:
@shyamk9 this looks great. Thanks for correcting the spelling mistake 👍

PR #167, opened 2017-11-03
32659_off_canvas _changes_committed with two of pending

HiirenP:
#32659 All changes are completed and ready for review **except Point No 3 & 4**. Girish will create a separate task for these two points.
Related to point no 2, (Dynamic size props) to work functionally, need to Remark media query from OffCanvas.scss file (Line number 68-75 and 111-117), as it prevents to set dynamic width for offCanvas Component. So you have to look for it if you can write some code related to it.


src/components/OffCanvas/OffCanvas.tsx
```diff
@@ -12,6 +12,13 @@ import OffCanvasContent from './OffCanvasContent';
 import * as baseTheme from './OffCanvas.scss';
 
 export type Mode = 'slide' | 'push' | 'reveal';
+export type WidthType = 'Small' | 'Medium' | 'Large' | string;
+
+export enum Width {
```
weegrom: @HiirenP the 'Small' | 'Medium' | 'Large' width options should apply a class, not a value. The display of the drawer needs to be defined and controlled from the scss file.

PR #173, opened 2017-12-14
fix for travis build - as it is not failing when unit tests are failed.

krunalm:


krunalm:
**Note:** Travis builds for this PR are failing as master branch have unit tests which are failing. Previously, when we have unit tests failing then Travis build is not failed, it was checking only linting and builds errors only, unit tests were optional. I think it should not be optional.

PR #177, opened 2018-02-27
Bugs/engage ui fix

harsimranb:
Fixed bugs and typescript issues.  updated Readme file.


src/components/MaskTextField/MaskTextField.tsx
```diff
@@ -407,7 +406,7 @@ class MaskTextField extends React.PureComponent<Props, State> {
       }
     }
     const passProps = copyProps ? copyProps : props;
-    return <TextField value={this.state.value} ref={(ref: Component<TextFieldProps>) => this.input = ref} {...passProps} />;
```
dheeraja00: @harsimranb I am just bit curious why `Component` had issue & `React.Component` don't

harsimranb: @dheeraja00 Reason was react library was being important twice, once * as React and the second {Compoennt}.  I removed the second one and had to change this to React.Component

PR #179, opened 2018-03-23
Features/33840 tablesorting

harsimranb:


src/components/Table/TableHead.tsx
```diff
@@ -0,0 +1,85 @@
...
+  // Accessibility id attribute, this will match with header attribute added in td
+  accessibilityId?: string;
+  // Accessibility scope attribute, this can be used instead of id
+  accessibilityScope?: Scope;
+  // Any custom class that user wants to add which could be for setting width or any custom styling
+  className?: string;
+  // If sorting is enabled, then clickhndler will get triggered on click
+  clickHandler(field: string): any;
+  // To set colspan value
+  colSpan?: number;
```
harsimranb: @dheeraja00 Rename to `columnSpan`

src/components/Table/TableHead.tsx
```diff
@@ -0,0 +1,85 @@
...
+  // Current sort order, it will be only filled if the current sorting field will be this
+  order?: string;
```
harsimranb: @dheeraja00 Let's make this an `enum`

src/components/Table/TableData.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { themr, ThemedComponentClass } from 'react-css-themr';
+
+import { TABLE } from '../ThemeIdentifiers';
+
+import * as baseTheme from './Table.scss';
+
+// All props start with `acc` are accessibility props
+export interface Props {
+  // Used for accessibility, match this with id set in <th>
+  accHeaders?: string;
```
harsimranb: @dheeraja00 rename to `accessibilityHeaders`

src/components/Table/TableData.tsx
```diff
@@ -0,0 +1,37 @@
...
+  colSpan?: number;
```
harsimranb: @dheeraja00 rename to `columnSpan`

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,257 @@
...
+              const thisSort: string = (sorting === 'all' || item.sort) && !item.noSort ? item.key : '';
+
+              return (
+                <TableHead
+                  key={item.key}
+                  sort={thisSort}
+                  style={item.style}
+                  className={item.className}
+                  order={field === item.key ? order.current : ''}
+                  clickHandler={this.sortData.bind(this)}>
```
harsimranb: @dheeraja00 Instead of bind, we can use arrow functions: `clickHandler={() => this.sortData()}`

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+    return classNames(
+      theme.table,
+      bordered && theme.bordered,
+      highlight && theme.highlight,
+      striped && theme.striped
+    );
+  }
+
+  /***
```
harsimranb: @dheeraja00 Remove commented code if no longer useful

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,257 @@
...
+  // Flag to indentify if table is sortable, if passed "all" then add sorting to all the columns
+  sorting?: boolean | string;
+  // Set greyed background for odd rows
+  striped?: boolean;
+  theme?: any;
+}
+
+export interface State {
+  data?: any;
+  sort?: any;
```
harsimranb: @dheeraja00 We need to explicity define an interface for this.  Check this out: https://www.typescriptlang.org/docs/handbook/interfaces.html

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,257 @@
...
+  // To make table bordered
+  bordered?: boolean;
+  // Column config, which renders the header
+  column?: any;
+  // Get the data & use it to populate tds
+  data?: any;
+  // On load by default it will be sorted by this field
+  defaultSortField?: string;
+  // Value could be 'asc' || 'desc'
+  defaultSortOrder?: string;
```
harsimranb: @dheeraja00 Use enum

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,257 @@
...
+  // Highlight rows on hover
+  highlight?: boolean;
+  // Make table responsive
+  responsive?: boolean;
+  // Flag to indentify if table is sortable, if passed "all" then add sorting to all the columns
+  sorting?: boolean | string;
```
harsimranb: @dheeraja00 Why string?

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,257 @@
...
+import TableRow from './TableRow';
+import TableData from './TableData';
+
+import * as baseTheme from './Table.scss';
+
+export interface Props {
+  // To make table bordered
+  bordered?: boolean;
+  // Column config, which renders the header
+  column?: any;
```
harsimranb: @dheeraja00 We need to explicity define an interface for this. Check this out: https://www.typescriptlang.org/docs/handbook/interfaces.html

example/src/App.tsx
```diff
@@ -116,6 +117,78 @@ class App extends React.Component<{}, State> {
...
+      injectHeader: To inject custom component in th
+    */
+    const columnConfig = [
+      {
+        label: '',
+        key: 'checked',
+        headerValue: false,
+        className: 'test',
+        style: { width: '80px' },
+        noSort: true,
```
harsimranb: @dheeraja00 Instead of `noSort: true`, let's just use `sort: false`.

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,257 @@
...
+ // Used certain flags which will help to add sorting for any specific fields
+  renderHeader() {
+    const { column, sorting } = this.props;
+    const { field, order } = this.state.sort;
+
+    return (
+      <TableHeader>
+        <TableRow>
+          {
+            column.map((item: any) => {
```
harsimranb: @dheeraja00  As stated above, we need an explicit interface, so it should be `(item: ColumnConfiguration) => { ... }`

example/src/App.tsx
```diff
@@ -181,14 +183,33 @@ class App extends React.Component<{}, State> {
...
+    /* 
+      Filtering table data
+      mode: Mode of filteration, right now we have only one which is search
+      search: If mode is search then it holds search config
+        event: When the search should fire, on click or on input
+        field: Name of field by which search should happen
+        placeholder: Filed placeholder
+        title: Title of field
+    */
+    const filterConfig = {
```
harsimranb: @dheeraja00 @weegrom The use case is that the filter textbox can be elsewhere on the page, in an entirely different segment.  In which case, it would be better if we just pass a filterQuery property to the table.

PR #180, opened 2018-03-23
Converted offcanvas to drawer & decoupled trigger with the component

dheeraja00:


src/components/Modal/Modal.tsx
```diff
@@ -1,115 +1,58 @@
...
-export interface ITrigger {
-  body: string;
-  animate: IAnimate;
-}
+export type Width = 'small' | 'medium' | 'large' | string;
 
 export interface Props {
+  active?: boolean;
+  activator: React.ReactNode;
+  children?: any;
```
harsimranb: @dheeraja00 We don't need children, React super class automatically gives us this.props.children.

src/components/Modal/Dialog.tsx
```diff
@@ -1,94 +1,72 @@
...
 
-export interface SizeStyle {
-  maxWidth: string;
-  width: string;
-}
+export type Width = 'small' | 'medium' | 'large' | string;
 
 export interface Props {
-  close?: boolean;
   children?: React.ReactNode;
```
harsimranb: @dheeraja00  no need for children in props

src/components/Modal/Dialog.tsx
```diff
@@ -1,94 +1,72 @@
...
-}
 
-export type Size = 'Small' | 'Medium' | 'Large' | number;
+import * as baseTheme from './Modal.scss';
 
-export interface SizeStyle {
-  maxWidth: string;
-  width: string;
-}
+export type Width = 'small' | 'medium' | 'large' | string;
```
harsimranb: @dheeraja00 It seems we're defining this propertyin multiple places.  let's reuse.

src/components/Modal/Modal.tsx
```diff
@@ -1,115 +1,58 @@
...
-  show: boolean;
+  active: boolean;
 }
 
 class Modal extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      show: false,
+      active: false,
```
harsimranb: @dheeraja00 good job changing the names of these properties!

PR #182, opened 2018-03-23
Modal refactored, decoupled trigger with modal component

dheeraja00:


src/components/Modal/Modal.tsx
```diff
@@ -1,130 +1,169 @@
...
-    }
+  render() {
+    return null;
+  }
+
+  // Render close button if its set true
+  renderCloseButton() {
+    const { closeButton, theme } = this.props;
+
+    return closeButton ? (<div className={theme.close}><Button plain onClick={this.closeModal.bind(this)} icon="cancel" /></div>) : null;
```
harsimranb: @dheeraja00 Instead of bind, use arrow functions

src/components/Modal/ModalHeader.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import { themr, ThemedComponentClass } from 'react-css-themr';
+
+import Heading from '../Heading';
+
+import { MODAL } from '../ThemeIdentifiers';
+
+import * as baseTheme from './Modal.scss';
+
+export type Width = 'small' | 'medium' | 'large' | string;
```
harsimranb: @dheeraja00 lets reuse the width type; it's defined at two places within modal.  (but drawer's width is separate from modal's width, keep that one there)

src/components/Modal/Modal.tsx
```diff
@@ -1,130 +1,169 @@
...
+  }
+
+  // Import keylistener if modal needs to be closed on pressing escape key
+  renderKeyListener() {
+    return this.props.closeOnEsc ? (<KeypressListener keyCode={Keys.ESCAPE} handler={this.triggerCloseEvent.bind(this)} />) : null;
```
harsimranb: @dheeraja00 Instead of bind, use arrow functions

src/components/Modal/Modal.tsx
```diff
@@ -1,130 +1,169 @@
...
-            width={width}
-          >
-            <div className={bodyClassNames}>
-              {children}
-            </div>
-          </Dialog>
+      <div
+        id={this.id}
+        className={modalWrapperClassName}
+        onClick={(closeOnBackgroud) ? this.triggerCloseEvent.bind(this) : undefined}>
```
harsimranb: @dheeraja00 Instead of bind, use arrow functions

src/components/Modal/Modal.tsx
dheeraja00: @harsimranb actually I am trying to keep it consistent with other function declaration, every function I have are function declaration, not the function expression. Let me know your thoughts

src/components/Modal/Modal.tsx
dheeraja00: Check my above reply on similar comment

src/components/Modal/tests/Modal.test.tsx
```diff
@@ -1,244 +0,0 @@
-import * as React from 'react';
```
harsimranb: @dheeraja00 You just deleted all the unit tests.  I hope you're planning to rewrite them.

PR #186, opened 2018-05-01
Features/35429 popover component

shyamk9:


docs/src/examples/DropdownExample/DropdownExampleFifth.tsx
```diff
@@ -0,0 +1,68 @@
...
+    this.setState({
+      active: !this.state.active
+    });
+  }
+
+  render() {
+    const items : DropdownItemProps[] = [
+      {
+        children: "Item 1",
+        header: false,
```
dheeraja00: @shyamk9 No need to add these false flag, remove all false flag from json & consider them as false in dropdown component if you don't receive the respective flag, add it only for true value

docs/src/examples/DropdownExample/DropdownExampleFifth.tsx
```diff
@@ -0,0 +1,68 @@
...
+        divider: false,
+        disabled: false
+      }
+    ]
+
+    return (
+      <div className={styles.example}>
+         <Dropdown
+          active={this.state.active}
+          DropdownItems={items}
```
dheeraja00: @shyamk9 Keep all keys in camel case, change it to dropdownItems

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+              divider={item.divider}
+              header={item.header}
+              children={item.children}
+            ></DropdownItem>
+        );
+      
+    return (
+      <WRAPPERCOMPONENT ref={this.setActivator}>
+        <div className={dropdownClassName} key={this.id}>
+          <Button onClick={toggle}>{children}</Button>
```
dheeraja00: @shyamk9 One point was decoupling the trigger from dropdown, you have done that but again here you have hardcoded the trigger to button, what if I want any anchor tag to be my trigger & what if I want to add any icon to my trigger

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+    const {
+      selectedIndex
+    } = this.state;
+
+    if (!active && !disabled) {
+      return;
+    }
+    
+    // Change Items on key up, down and tab
+    if (event.keyCode === Keys.UP_ARROW) {
```
dheeraja00: @shyamk9 When I press up & down arrow keys then it also move the whole body up & down that should be disabled

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+  children?: React.ReactNode;
+  disabled?: boolean;
+  divider?: boolean;
+  header?: boolean;
+  onClick?(): void;
+}
+
+export interface Props {
+  disabled?: boolean;
+  children?: string;
```
dheeraja00: @shyamk9 No need to add `children` prop in interface, its a default prop which comes with every component, so we don't have to give it a type

src/components/Dropdown/Dropdown.tsx
shyamk9: @dheeraja00 ,you can pass any react component in children. that will display as content in button.

src/components/Dropdown/Dropdown.tsx
shyamk9: ok, but i don't know how to do that.

src/components/Dropdown/Dropdown.tsx
dheeraja00: @shyamk9 Children is react's predefined prop, rename it to something else

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+export interface DropdownItemProps {
+  children?: React.ReactNode;
+  disabled?: boolean;
+  divider?: boolean;
+  header?: boolean;
+  onClick?(): void;
+}
+
+export interface Props {
```
dheeraja00: @shyamk9 There is no onClick or onSelect prop here, how user will know what item he has clicked & how can they decide what operation they want to perform onclick of an item, like:
1. if they want to navigate to specific route
2. Use dropdown like a filter or action menu, something similar to the dropdowns target process have

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -0,0 +1,61 @@
...
+    } else {
+      className = classNames(
+        baseTheme.dropdownItem,
+        active && baseTheme.active,
+        disabled && baseTheme.disabled
+      );
+    }
+
+    return (
+      <a id={this.id} className={className} onClick={onClick}>{children}</a>
```
dheeraja00: @shyamk9 It's not necessary that user will always want the anchor tag as there item, they can have button as well or simply div or any other tags, so make it dynamic

src/components/Dropdown/Dropdown.tsx
dheeraja00: @shyamk9 that's kind of a problem, suppose if I pass a button as children, then it will be like button inside button

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+import DropdownItem from './DropdownItem';
+import { classNames } from '@shopify/react-utilities/styles';
+import { DROPDOWN } from '../ThemeIdentifiers';
+import * as baseTheme from './Dropdown.scss';
+import { Keys } from '../../types';
+import Button from '../Button';
+
+export type Direction = 'up' | 'down' | 'left' | 'right';
+
+export interface DropdownItemProps {
```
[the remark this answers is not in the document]
dheeraja00: @shyamk9 all the values inside are looking same to me, just only active flag is missing from here, so I don't see an issue to import it from dropdownitem

src/components/Dropdown/Dropdown.tsx
dheeraja00: @shyamk9 you can check in react-strap, go to there github & take a look on how they have handled it, basically here you are not looking for an event for any specific element, it seems you are checking that in document level, that could be the one reason, but still check there repo

src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
shyamk9: @dheeraja00 if i use dropdown item props directly the user of dropdown component can be confused as there are active in props that is of not use is it ok ?

docs/src/examples/DropdownExample/DropdownExampleFifth.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import { Dropdown, DropdownItemProps } from '../../../../src/components/';
```
ankitsaraogi: last '/' after components is not required, please remove it from everywhere.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  private activatorContainer: HTMLElement | null;
+
+  private id = getUniqueID();
+
+  componentWillReceiveProps(newProps: any) {
+    const { active } = this.props;
+
+    if (active && !newProps.active && typeof newProps.onClose !== 'undefined') {
```
ankitsaraogi: @shyamk9: I think in if check you meant, 'newProps.onClose !== undefined', right now it will check for string undefined. Please correct it.

dheeraja00: @ankitsaraogi good find 👍

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+    const DropdownItemComponents = DropdownItems.map((item,index) => 
+            <DropdownItem 
+              key={index}
+              active={selectedIndex === index} 
+              disabled={item.disabled}
+              divider={item.divider}
+              header={item.header}
+              children={item.children}
+            ></DropdownItem>
+        );
```
ankitsaraogi: @shyamk9 : Make this as self ending tag, as it doesn't have anything inside the <DropdownItem>

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+    super(props);
+    this.state = { 
+      selectedIndex: 0 
+    };
+  }
+
+  private activatorContainer: HTMLElement | null;
+
+  private id = getUniqueID();
+
```
ankitsaraogi: @shyamk9 : why are we keeping id & activatorContainer as global variable inside class, can we make it attach to 'this' and use it

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+    } = this.props;
+    
+    const {
+      selectedIndex
+    } = this.state;
+
+    
+    const dropdownClassName = classNames (
+      typeof direction == 'undefined' || direction === "down" ? baseTheme.dropdown 
+        : direction === "up" ? baseTheme.dropup
```
ankitsaraogi: @shyamk9 : same undefined check, please correct.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,214 @@
...
+    focusableActivator.setAttribute('aria-controls', id);
+    focusableActivator.setAttribute('aria-owns', id);
+    focusableActivator.setAttribute('aria-haspopup', 'true');
+    focusableActivator.setAttribute('aria-expanded', String(this.props.active));
+  }
+
+  @autobind
+  private setActivator(node: HTMLElement | null) {
+    if (node == null) {
+      this.activatorContainer = null;
```
ankitsaraogi: @shyamk9 : put a '===' check, instead of '== null' check.

src/components/Popover/Popover.tsx
```diff
@@ -1,121 +0,0 @@
...
-  sectioned?: boolean;
-  onClose?(source: CloseSource): void;
-}
-
-export interface State {
-  activatorFocused: boolean;
-}
-
-const getUniqueID = createUniqueIDFactory('Popover');
-
```
ankitsaraogi: We can keep this inside class, instead of declaring outside.

src/components/Dropdown/Dropdown.tsx
shyamk9: @ankitsaraogi but it is not working, if i do newProps.onClose !== undefined then it is giving me Dropdown.tsx:28 Uncaught TypeError: newProps.onClose is not a function error when onClose prop is not passed

dheeraja00: @shyamk9 can you console the `newProps.onClose` and check what it gives, also I think you don't need to check its type, it should work fine like this: `if (active && !newProps.active &&  newProps.onClose)`

shyamk9:
@dheeraja00 and @ankitsaraogi I have done require review changes please check it and merge the pull request.

dheeraja00:
@shyamk9 Dropdown arrow is not coming poperly, & previously you had issue with only right position, but right now it seems none of the position working properly

PR #187, opened 2018-05-01
Features/35426 description list component

ishani-m:


example/src/App.tsx
```diff
@@ -600,6 +608,55 @@ class App extends React.Component<{}, State> {
             <Item>Second item</Item>
             <Item>Third Item</Item>
           </List>
+          <Heading>Description List</Heading> 
+          <DescriptionList type="default" style="Inline"           
+            items={[
```
dheeraja00: @ishani-m Here you can only pass the textual data, but there should be a way to pass different component to both term & description, it should support to show simple text or any component

PR #188, opened 2018-05-03
Features/35424 tabs component

HiirenP:


src/components/Tab/Tab.tsx
```diff
@@ -0,0 +1,35 @@
...
+export interface Props {
+  // Description of tab (like title only, title with icon, or any other react component)
+  tabDescription?: React.ReactNode;
+  // Unique identity for tab
+  tabId: string;
+  // Identity of selected/currently active tab
+  activeTabId?: string;
+  // Set theme for tab
+  theme?: any;
+  children?: any;
```
dheeraja00: @HiirenP No need to define children prop as its a pre-defined prop which comes with every component, so we don't have to define its type

src/components/Tab/TabPanel.tsx
```diff
@@ -0,0 +1,121 @@
...
+  // left - tabs alignment starting from left side
+  // center - tabs alignment will be in center
+  // right - tabs alignment starting from right side
+*/
+export type Alignment = 'left' | 'center' | 'right';
+
+// All prototypes type
+export interface Props {
+  // Children are the tabs and it's content
+  children?: any;
```
dheeraja00: @HiirenP Check my above comment

src/components/Tab/TabPanel.tsx
```diff
@@ -0,0 +1,121 @@
...
+  // Position prop defines the tab's location with reference to it's content
+  position: Position;
+  // Alignment prop defines the tab's Alignment in tab panel
+  alignment: Alignment;
+  // For Default active tab
+  defaulttabid?: string;
```
dheeraja00: @HiirenP Use camel case for all the keys, so it should be `defaultTabId`

src/components/Tab/TabPanel.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  // This will render single/multiple tabs into panel with props
+  renderTabs() {
+    const { activeTabId } = this.state;
+    // Iterate through a list of tabs and bind onclick event respectively
+    return React.Children.map(this.props.children, (child: React.ReactElement<any>, index) => {
+      return React.cloneElement(child, {
+        activeTabId,
+        onClick: this.onTabClick.bind(this, this.props.children[index].props.tabId),
```
dheeraja00: @HiirenP 
1. Using `bind` can cause the performance issue, so declare `onTabClick` as arrow function this way you don't have to use bind
2. You don't have to use like `this.props.children[index].props.tabId` I believe this can be changed to `child.props.tabId`

src/components/Tab/TabPanel.tsx
HiirenP: I was forced to do this change for only prop because i was getting a warning message something like this. "React does not recognize the `defaultTabId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `defaulttabid` instead".

spent some time to research on this but not found any solution can you please help me in this?

src/components/Tab/TabPanel.tsx
HiirenP: resolved!

src/components/Tab/TabPanel.tsx
HiirenP: resolved!

src/components/Tab/Tab.tsx
HiirenP: resolved!

src/components/Tab/TabPanel.tsx
dheeraja00: This is a weird issue, I will have a look

dheeraja00:
@HiirenP if you have taken care of all the comments then please try to resolve travis build

HiirenP:
Regarding Style/Class props i have added it into tab component earlier, but after confirming the requirement, i had removed it, but anyways i have added it again.

PR #190, opened 2018-05-04
Features/35427 alert component

ishani-m:


example/src/App.tsx
```diff
@@ -428,6 +432,22 @@ class App extends React.Component<{}, State> {
         <br />
         <Caption style={{ color: 'red' }}>This is Caption</Caption>
         <br />
+        <Heading>Alert</Heading>
+        <Alert>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt.
```
ankitsaraogi: Indent the code properly, i.e. give tab to the text.

example/src/App.tsx
```diff
@@ -629,7 +652,10 @@ class App extends React.Component<{}, State> {
             autoSuggest
             moreInfoComponent={<Button children="ranmal" />}
           />
-          <ValidatedForm>
+          <ValidatedForm
+            onSubmitError={(value: [any], error: Error) => console.log('value:', value, 'error:', error)}
+            onSubmit={(value: [any]) => console.log('Submit Value:', value)}
+          >
```
ankitsaraogi: Do we need console.log here, if no, then remove them.

example/src/App.tsx
```diff
@@ -665,9 +690,23 @@ class App extends React.Component<{}, State> {
                   { required: true, message: 'App Description is required.' },
                 ]}
               />
+
+              <ValidatedSelectField
+                id="appCity"
+                required={true}
```
ankitsaraogi: just pass 'required', no need to pass 'required={true}'

src/components/Alert/Alert.scss
```diff
@@ -0,0 +1,73 @@
+@import 'foundation/spacing';
+@import 'shared/typography';
+
+.alert {
+    position:relative;
+    margin-bottom:20px;
+    padding:15px 29px 15px 15px;
```
ankitsaraogi: I feel instead of '29px' lets keep it '30px', so that all pixels are near rounded.

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </div>}
+        </React.Fragment>
+      );
+    }
+    closeAlert = () => {
+      this.setState({
+        isActive: false,
+      });
+    }
+  }
```
ankitsaraogi: Its always nice to write render as the last function of any component. So, please move render at the last function in class.

src/components/Picker/Picker.tsx
```diff
@@ -105,6 +106,12 @@ class Picker extends React.Component<Props, State> {
       number: 0,
     };
   }
+
+  componentWillReceiveProps(newProps: Props) {
+    if (newProps.source.length !== this.props.source.length) {
+      this.setState({ itemsList: newProps.source });
+    }
```
ankitsaraogi: Here logic looks wrong, what will happen if source is different object array but of same length as prev props source.
We should check if nextProps.source array is different of props.source array.

PR #191, opened 2018-05-09
Features/35425 accordion component

shyamk9:


docs/src/examples/AccordionExample/AccordionExampleFirst.tsx
```diff
@@ -0,0 +1,69 @@
...
+class AccordionExample extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      Item1: false,
+      Item2: false,
+      Item3: false
+    };
+
```
ankitsaraogi: I feel that we can take items as an array, instead of defining each items separately. If there are 10 items, we don't want to define 10 items in state.

docs/src/examples/AccordionExample/AccordionExampleFirst.tsx
```diff
@@ -0,0 +1,69 @@
...
+      Item2: !this.state.Item2
+    });
+  }
+
+  private Item3Toggle() {
+    this.setState({
+      Item3: !this.state.Item3
+    });
+  }
+
```
ankitsaraogi: Where are you calling all the toggle functions, this looks wrong.

ankitsaraogi: Also is there a way, we can always open only 1 item, i.e. say item1 is open, and then when I click item2, item1 gets close and item2 gets open. We should have that functionality as well which can be driven via property, if anyone wants only 1 to be open at any given time.

PR #194, opened 2018-05-18
Treeview component

dheeraja00:


src/components/TreeView/TreeNode.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  renderNode = (): any => {
+    const { label = '' } = this.props;
+
+    return label;
+  }
+
+  render() {
+    // If component prop is not present then check for label & render that
+    const { component = this.renderNode } = this.props;
```
harsimranb: @dheeraja00 Can you help me understand the syntax here? `{ component = this.renderNode }` if component is null, then it sets it to renderNode, else it doesn't.  Is this correct?

src/components/TreeView/TreeView.tsx
```diff
@@ -0,0 +1,134 @@
...
+          this.setState({ source });
+          return true;
+        }
+      }
+    });
+  }
+
+  // Set child nodes active or inactive status
+  toggleChildNode = (source: SourceData[], id: number) => {
+    return source.some((item: SourceData): any => {
```
harsimranb: I believe you don't have to specify `: any` here.  If you do, i think it should be a boolean.  Let's not use `any` unless we absolutely have to.

src/components/TreeView/TreeView.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+      // Recursive if child node contains another child node
+      if (item.children) {
+        this.toggleChildNode(item.children, id);
+      }
+    });
+  }
+
+  // Render single node, also iterate through its children & render those as well if its parent active status is true
+  renderNode = (item: SourceData): any => {
```
harsimranb: Same with `:any` here as before.

src/components/TreeView/interface.ts
```diff
@@ -0,0 +1,11 @@
+import { ReactElement } from 'react';
+
+export interface SourceData {
```
harsimranb: Good work!

src/components/TreeView/interface.ts
```diff
@@ -0,0 +1,11 @@
...
+  id: number;
+  label?: string;
+  component?(): ReactElement<any>;
```
harsimranb: @dheeraja00 Why is it better to make `component` a function?

dheeraja00:
@harsimranb I have made the required changes please check

PR #195, opened 2018-05-18
Dropdown positioning changes

dheeraja00:


src/components/Dropdown/Dropdown.tsx
```diff
@@ -34,6 +34,8 @@ export class Dropdown extends React.PureComponent<Props, State> {
   private getUniqueID = createUniqueIDFactory('Dropdown');
   private activatorContainer: HTMLElement | null;
   private id = this.getUniqueID();
+  private dropdownEle: any = null;
```
harsimranb: @dheeraja00 As mentioned before, we want to write out entire names: `dropdownElement`  Also, we want to avoid `: any`.  A simple google search returned https://stackoverflow.com/questions/14742194/declaring-an-htmlelement-typescript and `HTMLElement is the type we can use.`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -34,6 +34,8 @@ export class Dropdown extends React.PureComponent<Props, State> {
...
+  private dropdownOffset: any = { height: 0, width: 0 };
```
harsimranb: @dheeraja00 Again, not a fan of `: any`.  I think you can remove `: any` and typescript might infer the type.  Even though this is private, linting and ts is not of much help if we use any.  It ignores type checking for that var, defeating the whole point.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -106,35 +115,41 @@ export class Dropdown extends React.PureComponent<Props, State> {
 
     const activatorComp = anchorEl;
     let activatorRect: ClientRect | DOMRect;
-    let leftCord : number = 0;
+    let dropdownPostion: any = {};
```
harsimranb: `: any` is no good.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -106,35 +115,41 @@ export class Dropdown extends React.PureComponent<Props, State> {
...
+    // let leftCord : number = 0;
+    // const dropdownPosition: any = {};
     if (activatorComp != null) {
       activatorRect = activatorComp.getBoundingClientRect();
-      if (direction === 'up' || direction === 'down') {
-        leftCord = activatorRect.left + (activatorRect.width / 2) - 50 /* Menu Width / 2 */;
+      if (direction === 'up') {
+        dropdownPostion = { top: - this.dropdownOffset.height - activatorRect.height };
       } else if (direction === 'left') {
-        leftCord = activatorRect.left - 100 /* Menu Width */;
```
harsimranb: Good job removing hardcoded number!

src/components/Dropdown/Dropdown.tsx
dheeraja00: @harsimranb sure I will avoid use of `any` anywhere, and will make changes to every places where I have used it

dheeraja00:
@harsimranb I have taken care of all above posted comments

harsimranb:
@dheeraja00 Looks like unit tests are failing.  Check Travis build log.

dheeraja00:
@harsimranb Travis build is fixed

dheeraja00:
@harsimranb Since travis build is fixed I have merged the branch, still not deleting it so if you need you can check the changes

PR #197, opened 2018-05-24
Features/35592 side navigation component

ishani-m:


src/components/SideNavigation/SideNavigation.tsx
```diff
@@ -0,0 +1,140 @@
...
+import Accordion from '../Accordion';
+import {AccordionItemProps} from '../Accordion';
+import * as baseTheme from './SideNavigation.scss';
+
+export interface INavigationData {
+    id?:number;
+    label?: React.ReactNode;
+    icon?: React.ReactNode;
+    children?: React.ReactNode;
+    action?(): void;
```
dheeraja00: @ishani-m Action can also accept any argument & then send it back in callback function so please add that & make is optional its not necessary that user will want that everytime

src/components/SideNavigation/SideNavigation.tsx
```diff
@@ -0,0 +1,140 @@
...
+        if (bodyElement !== null) {
+            bodyElement.className +=  this.state.drawer ? ' ' + this.props.theme.container : '';
+        }
+    }
+
+    render() {
+        const { source, theme, accordian } = this.props;
+        let actDrawerId = this.state.activeDrawerId;
+        const rootElement = document.getElementById('root');
+        const iconClass = theme.icon;
```
dheeraja00: @ishani-m Please destructure theme values as well instead of creating every single variables, it will be something like: `const { collapseIcon, navDivider, li: liClass } = theme;`, so if you want to rename the variable you can write like `li: liClass`

src/components/SideNavigation/SideNavigation.tsx
```diff
@@ -107,27 +105,29 @@ class SideNavigation extends React.Component<Props, State> {
                 </p>
             );
         });
+        let collapseIcon:any = actDrawerId == "fullContent" ? "chevronLeft" : "chevronRight";
```
dheeraja00: @ishani-m Please use proper type, don't ever use `any`

src/components/SideNavigation/SideNavigation.tsx
```diff
@@ -14,87 +13,86 @@ export interface INavigationData {
...
-        const liClass = theme.li;
-        const childLiClass = theme.childLi;
-        let accState = accordian;
         if (rootElement !== null) {
-            rootElement.className = this.state.drawer ? (theme.container) : '';
+            rootElement.className = drawerOpen ? (theme.container) : '';
             rootElement.className = rootElement.className + ' ' + (actDrawerId == "collapsedContent" ? this.props.theme.rootCollapse : '')
         }
 
         const fullContentMarkup = source.map(function(full:any){
```
dheeraja00: @ishani-m Please use arrow functions in every places

PR #219, opened 2018-09-27
Features/36681 sticky footer component

HiirenP:


docs/src/reducers/componentState/StickyState.ts
```diff
@@ -0,0 +1,41 @@
...
+    {
+      name: 'children',
+      type: 'any',
+      desc: 'Elements to display inside Sticky.',
+    }, {
+      name: 'position',
+      type: 'Position',
+      desc: 'Position prop allow user to choose StickyHeader or StickyFooter',
+    }, {
+      name: 'style',
```
dheeraja00: @HiirenP We are using `componentStyle` key for style props

src/components/Sticky/Sticky.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import { themr, ThemedComponentClass } from 'react-css-themr';
+import { STICKY } from '../ThemeIdentifiers';
+import { classNames } from '@shopify/react-utilities/styles';
+import * as baseTheme from './Sticky.scss';
+
+export type Position = 'top' | 'bottom';
```
dheeraja00: @HiirenP Can you also please add left & right position as well

src/components/Sticky/Sticky.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export interface Props {
```
dheeraja00: @HiirenP also please add `componentClass` in the prop, its possible that we would want to pass any class as a prop from emgage-frontend, so it should include that class here

src/components/Sticky/Sticky.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children?: React.ReactNode;
+  position: Position;
+  style?: React.CSSProperties;
+  // Theme to be injected via css-themr.
+  theme?: any;
+}
+
+class Sticky extends React.PureComponent<Props, any> {
```
dheeraja00: @HiirenP since it's a stateless component, don't use class check other component like description or button

docs/src/reducers/componentState/StickyState.ts
HiirenP: resolved

src/components/Sticky/Sticky.tsx
HiirenP: resolved

src/components/Sticky/Sticky.tsx
HiirenP: But it will work for particular div only. and position is for like top for stickyHeader and bottom for stickyFooter. and so as this will occupy 100% width and most cases due to Responsiveness of components and elements in emgage-ui, i guess this will not gonna work.

src/components/Sticky/Sticky.tsx
HiirenP: resolve

PR #220, opened 2018-10-02
feat: process component

HiirenP:


src/components/Process/Process.tsx
```diff
@@ -0,0 +1,176 @@
...
+  // To apply any theme to component via themer.
+  theme?: any;
+  // To apply custom styling.
+  componentStyle?: React.CSSProperties;
+    // Custom class
+  componentClass?: string;
+}
+
+export interface State {
+  showPreviousBtn: boolean;
```
dheeraja00: @HiirenP Always use the full name, like `showPreviousButton` instead of `Btn`, do this change in all the places of this component

src/components/Process/Process.tsx
```diff
@@ -0,0 +1,176 @@
+import * as React from 'react';
+import { themr, ThemedComponentClass } from 'react-css-themr';
+import { classNames } from '@shopify/react-utilities/styles';
+import { PROCESS } from '../ThemeIdentifiers';
+import * as baseTheme from './Process.scss';
+
+export interface NavState {
```
dheeraja00: @HiirenP rename it to NavigationState

src/components/Process/Process.tsx
```diff
@@ -0,0 +1,176 @@
...
+  indx: number;
+  styles: string[];
+}
+
+export interface Step {
+  name: string;
+  component: React.ReactNode;
```
dheeraja00: @HiirenP user can also pass only the string, its not necessary to pass node always, so add `| string` type as well

src/components/Process/Process.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { classNames } from '@shopify/react-utilities/styles';
+import { PROCESS } from '../ThemeIdentifiers';
+import * as baseTheme from './Process.scss';
+
+export interface NavState {
+  indx: number;
+  styles: string[];
+}
+
+export interface Step {
```
dheeraja00: @HiirenP Add onClick callback as well, make it optional & if user passed it by default return the `event` in callback, so it will be like: `onClick?(returnValue: any): void`

dheeraja00: But remember this won't work for the inactive state, so if there are 3 steps & 1 step is done & if user tries to click on 3rd step directly then you should not call this callback function

dheeraja00:
@HiirenP I ran the process component & its difficult to know which process is completed & you seem to have added that next or previous button which is not required, check it in video or the image, we won't be having any buttons. You are also showing the digit list not sure why

PR #235, opened 2018-11-28
Features/36967 checkbox on parent checked all child when checked

HiirenP:


src/components/Table/Table.tsx
```diff
@@ -357,7 +357,8 @@ class Table extends React.Component<Props, State> {
 
   // Function to add checkbox in header as well
   addHeaderCheckbox = (): React.ReactElement<any> => {
-    return <TableHead componentStyle={{ width: 'auto' }}><Checkbox label="" checked={this.state.allRowChecked} onChange={this.toggleAllRowSelection} /></TableHead>;
+    const isAllChecked = ((this.state.totalRowCount - this.state.selectedRows.length) > 0 &&  this.state.totalRowCount > 0) ? true : false;
+    return <TableHead componentStyle={{ width: 'auto' }}><Checkbox label="" checked={isAllChecked && this.state.selectedRows.length > 0 ? true : (this.state.selectedRows.length > 0 && this.state.totalRowCount > 0 && this.state.totalRowCount === this.state.selectedRows.length) ? true : this.state.allRowChecked} indeterminante={isAllChecked} onChange={this.toggleAllRowSelection} /></TableHead>;
```
dheeraja00: @HiirenP Instead of defining the whole condition directly in jsx, its better o store the result in constant & then use it.

src/components/Table/Table.tsx
```diff
@@ -452,42 +451,92 @@ class Table extends React.Component<Props, State> {
...
+        if (this.props.nestedChildData[key].component.props.data.length > 0) {
+          for (const index in this.props.nestedChildData[key].component.props.data) {
+            allChildData.push(this.props.nestedChildData[key].component.props.data[index]);
+          }
+        }
+      }
+    }
+
+    allChildData.push(...this.state.data);
+    const allData = [...new Map(allChildData.map((o: any) => [o.id, o])).values()];
```
dheeraja00: @HiirenP don't use short forms like this, use childData as param or some valid name instead of just `o`

src/components/Table/Table.tsx
```diff
@@ -77,17 +76,18 @@ export interface State {
...
+
 class Table extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
-
     this.state = this.getInitialState();
   }
 
   componentWillMount() {
-    // If user asked to sort the data by default then call the sortdata before rendering
```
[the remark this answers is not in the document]
HiirenP: By mistake ! may be in process of removing comments and remarks! Will be careful.

PR #242, opened 2018-12-24
Feature/engage ui style

dheeraja00:
TP: 37588, 37777, 37796


example/src/App.tsx
```diff
@@ -1004,7 +1004,7 @@ class App extends React.Component<{}, State> {
             active={this.state.outterDrawer}
             activeContentId={this.state.outterDrawerId}
             mode="slide"
-            componentWidth="large"
+            componentWidth="530px"
             componentLabel={this.state.outterDrawerLabel}
```
ankitsaraogi: Can we make '530'as a constant, and use it that way.

src/components/Drawer/Drawer.tsx
```diff
@@ -176,7 +192,7 @@ class Drawer extends React.PureComponent<Props, never> {
   }
 
   renderLayer() {
-    const { active, mode, componentLabel, componentWidth, theme } = this.props;
+    const { active, mode, componentLabel, componentWidth, theme,  } = this.props;
     const containerClassName = this.getContainerClassName();
```
ankitsaraogi: Is the extra comma intentional ?

src/components/Select/Select.tsx
```diff
@@ -120,7 +120,7 @@ const select = ({
           id={componentId}
           name={name ? name : 'select'}
           value={value}
-          className={theme.input}
+          className={theme.select}
           disabled={disabled}
```
ankitsaraogi: Are we supporting 2 kind of 'select', I see one as 'Select' & another one as 'select'.

example/src/App.tsx
dheeraja00: @ankitsaraogi user can pass value in px or em or rem, that's why I have kept it like that

src/components/Drawer/Drawer.tsx
dheeraja00: @ankitsaraogi its in tsconfig that every object should end with comma, that's why its there

src/components/Select/Select.tsx
dheeraja00: @ankitsaraogi yes that's kinda confusing, one style is in delicious theme & other is in component folder, so I am not changing any of them as that can cause any conflict

PR #248, opened 2019-01-04
Features/37585 card component styling

HiirenP:


example/src/App.tsx
```diff
@@ -1299,6 +1303,24 @@ class App extends React.Component<{}, State> {
             resizable
             // onChange={this.valueUpdater('appTextCounter')}
           />
+                                <Card>
+                                    <CardBody componentTitle="Online store dashboard - Card" sectioned>
```
dheeraja00: @HiirenP What's the use of componentTitle here? now since we already have `CardHeader`, we can use that to show the title right?

src/components/Card/Card.scss
```diff
@@ -23,7 +31,18 @@
 }
 
 .header {
-  padding: spacing(loose) spacing(loose) 0;
+  padding: .75rem 1.25rem;
```
dheeraja00: @HiirenP don't ever use padding or margin in any of the components directly, we want to make all the components generic. In some places we might need less padding/margin and in some more, so its better to not give this values at all. This values will be passed to component using `componentStyle` or `componentClass`

src/components/Card/Card.scss
```diff
@@ -45,5 +64,14 @@
 .footer {
   display: flex;
   justify-content: flex-end;
-  padding: 0 spacing(loose) spacing(loose);
+  // padding: 0 spacing(loose) spacing(loose);
+}
+
+.cardTitle {
+  font-size: 24px;
+  font-family: OpenSans-Semibold;
```
dheeraja00: @HiirenP I am not sure if font-family is needed here, if yes then its better to add that in delicious theme itself in a common class, because it might be needed in many more places

src/components/Card/Card.scss
```diff
@@ -45,5 +64,14 @@
...
+  letter-spacing: 0;
+  color: #212121;
```
dheeraja00: @HiirenP like padding & margin also don't give color values to any component

src/components/Card/CardFooter.tsx
```diff
@@ -0,0 +1,59 @@
...
+    : null;
+
+  const secondaryFooterActionMarkup = secondaryFooterAction
+    ? buttonFrom(secondaryFooterAction)
+    : null;
+
+  const footerMarkup = primaryFooterActionMarkup || secondaryFooterActionMarkup
+    ? (
+      <div className={theme.footer}>
+        <ButtonGroup>
```
dheeraja00: @HiirenP its not necessary that footer will have buttons, user can put anything there, keep it generic so it can accept any type of element or component, its just should be separate from header & body

HiirenP: Yes you are correct but this is for predefined buttons(like ok, cancel / save, cancel etc..) and this will be rendered only if we give them a primary or secondary action via props, otherwise it any react component as child will be rendered. which is in else condition. and this was coded like this for card component before. should i still remove it?

example/src/App.tsx
HiirenP: I have seen at many places, where header is used as heading of card in general and componentTitle is used as subTitle of card body and that can be multiple as no. of sections in card body. but should i still remove it according to you if will not be useful in future?

src/components/Card/CardFooter.tsx
HiirenP: @dheeraja00  Please see comments above i have added

example/src/App.tsx
dheeraja00: @HiirenP yeah I don't see the use of it, lets remove it

src/components/Card/CardFooter.tsx
dheeraja00: > Yes you are correct but this is for predefined buttons(like ok, cancel / save, cancel etc..) and this will be rendered only if we give them a primary or secondary action via props, otherwise it any react component as child will be rendered. which is in else condition. and this was coded like this for card component before. should i still remove it?

Hmm, yeah I think let's remove it because this will create more complications, like what should happen on click of those buttons, if a user wants to add an extra style to those buttons or call specific functions on click or disable it. So its better to remove it

HiirenP:
@dheeraja00  Resolved

PR #270, opened 2019-03-25
fix: id:39218 fix tab component issue

HirenTariwala:


docs/src/examples/TabExample/TabExampleFirst.tsx
```diff
@@ -23,7 +23,7 @@ class TabExample extends React.Component {
   render() {
     return (
       <div className={styles.example}>
-        <TabPanel position={'top'} alignment={'center'} defaultTabId={this.state.activeTabId} >
+        <TabPanel position={'top'} alignment={'center'} defaulttabid={this.state.activeTabId} >
```
dheeraja00: @HirenTariwala Instead of changing the name, I would suggest to find the proper solution, so we can use camelcase everywhere & we don't have to change the name

HirenTariwala: @dheeraja00 right now I am following suggestion given by warning itself other one suggestion is just removed defultTabId . let me check is there any other solution or not

PR #357, opened 2019-06-27
fix: minor change

dheeraja00:


weegrom:
@dheeraja00 why are you changing all the "icon" class names to "customIcon"? I've noticed you're adding "custom..." classes a lot of places and don't see a reason for it. Themr allows us to override the styling of any class within a component, and css-modules keeps things specific to said component... so I don't see the need for this change. I actually see it as confusing. I'd like to hear your reasoning behind making the changes, as it may be a good reason the I'm simply unaware of.

dheeraja00:
@weegrom I knew I will probably hear this from you, so let's take an example of Badge component, we can add different Icon component in Badge and one of them which is currently added is refreshing icon `<Icon source="refresh" componentColor="inkLighter" theme={theme} />` now if badge component's css has an .icon class which it has:
```
.iconWrapper {
  height: $icon-size;
  width: $icon-size;
  display: inline-block;
  vertical-align: top;
  margin: $icon-spacing $icon-spacing 0 ($icon-spacing - $horizontal-padding + rem(2px));
}
```
so if we just pass it as `.icon` then this create different issues specially that margin, actually you can simply rename `iconWrapper` to `icon` and you will see what I am trying to say

weegrom:
@dheeraja00 ha ha... woah... you went a little overboard in passing the theme prop within components. You're gonna run into the same problem again with what's been done... all I need to do is add an Icon in the emgage-frontend app and apply a theme where I use the .icon class, and all those Icon components within emgage-ui that you're not passing the theme prop to are going to take on that styling :O. This warranted a more thoughtful approach... passing the theme prop to the components used within the emgage-frontend app made good sense in many cases... however passing it to all child components within the engage-ui components went too deep.

PR #391, opened 2019-10-11
Feature/38327 button add props

webdevlapani:


dheeraja00:
@webdevlapani I can't understand the purpose of togglebuttongroup component, can't we use buttongroup component, because I think you are looking for the way to show those two buttons for the bit widgets

PR #403, opened 2020-01-09
fix: webpackv4 support

harsimranb:


harsimranb:
@dheeraja00 Please confirm you can compile and run this branch locally before I merge.

harsimranb:
@weegrom @dheeraja00  We need to get this tested and reviewed.

dheeraja00:
@harsimranb @weegrom this is looking ok, I was getting the error related to html-webpack-plugin, so once I updated its version & now it's working fine. I also tested most of the components which we are using and it seems to be working well. So I am going to merge this PR & than will test it in emgage-frontend

dheeraja00:
@harsimranb I think this should have been auto publish but it has not

harsimranb:
@dheeraja00 Yeah, weird - it didn't auto publish.  Can you manually publish it when you have a chance?  thanks

PR #433, opened 2020-04-14
fix: 44330, 44023: Popover Z-index, language icon overlap fix

Farhan-emgage:
Bug fixing for popover overlaping with header and laguage icon overlapping with sticky footer

### DIFF
--- a/src/components/ComboBox/ComboBox.scss
+++ b/src/components/ComboBox/ComboBox.scss
@@ -1,47 +1,43 @@
 $border-color: #eceeef;
 $backrgound-color: #F9FAFB;
-
 .comboboxContainer {
-  position: relative;
+    position: relative;
 }
 
 .accordionItemHeader {
-  display: flex;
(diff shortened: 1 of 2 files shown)


weegrom:
@Farhan-emgage you need to be making requested changes within one day of them being requested. This PR should not still be sitting open.

src/components/ComboBox/ComboBox.scss
```diff
@@ -1,47 +1,50 @@
 $border-color: #eceeef;
 $backrgound-color: #F9FAFB;
 
+$stacking-order: (
+  input: 20,
```
weegrom, line 5: @Farhan-emgage typically we don't map for variables that don't exist in the current context. For example, input, backdrop, and spinner are not used in the file. Only icon in used, so that's the only one you need to include. I'll approve this PR so the fix get rolled out this sprint, however please remove the unused map items at some point. Thanks

PR #444, opened 2020-04-30
Feature/multicheckboxfacets

ankitsaraogi:


example/src/App.tsx
```diff
@@ -137,6 +138,7 @@ interface State {
   treeAnchor?: any;
   ChoiceListSelected: any[];
   paginationCurrent: number;
+  multipleCheckboxFacetsOptions: anu[];
```
HirenTariwala, line 141: anu is any interface or it is by mistake of any ?

PR #475, opened 2020-06-11
44870 update react-router and redux

volodymyr-code:


ankitsaraogi:
@volodymyr-code : Can you please confirm if all the code that we removing isnt getting used anywhere.
Also please check pipeline

volodymyr-code:
> @volodymyr-code : Can you please confirm if all the code that we removing isnt getting used anywhere.
> Also please check pipeline

in my opinion, all works fine, confirm
pipeline depends on merging builder and ui PRs

volodymyr-code:
I can run build locally and leave output here

PR #479, opened 2020-06-15
fix: 44550 Group AddMember member not displayed in dropdown bug fixed

romesh87:
https://atera.tpondemand.com/entity/44550-fe-group-add-member-greater-member
bug fixed

### DIFF
--- a/src/components/Picker/Picker.tsx
+++ b/src/components/Picker/Picker.tsx
@@ -236,7 +236,7 @@ class Picker extends React.Component<Props, State> {
         const addedItem = this.state.chipListState.slice(number, number + 1);
         const addedItemObj = Object.assign({}, addedItem[0], { tabIndex: -1 });
         const chipListState = existingChipList.slice(0, number).concat(existingChipList.slice(number + 1));
-        const itemsList = this.state.itemsList.concat([addedItemObj]);
+        const itemsList = [addedItemObj, ...this.state.itemsList];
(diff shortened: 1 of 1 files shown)


src/components/Picker/Picker.tsx
```diff
@@ -236,7 +236,7 @@ class Picker extends React.Component<Props, State> {
         const addedItem = this.state.chipListState.slice(number, number + 1);
         const addedItemObj = Object.assign({}, addedItem[0], { tabIndex: -1 });
         const chipListState = existingChipList.slice(0, number).concat(existingChipList.slice(number + 1));
-        const itemsList = this.state.itemsList.concat([addedItemObj]);
+        const itemsList = [addedItemObj, ...this.state.itemsList];
```
CaptainSadaf, line 239: @romesh87 Does it solve the drop down display issue by adding item with spread operator, I think the later solution is adding new item in the start of of itemsList object rather than in the last?
Could you please let me know the actual difference of these two lines relate to group added member display issue?

romesh87: @CaptainSadaf  Correct, the issue was that the deleted item was appended at the end of the  autosuggestion array  and so it was not visible immediately in the dropdown. Now it is the first item in the array and can be seen immediately.

CaptainSadaf: @romesh87 Alright! Thanks

PR #480, opened 2020-06-16
fix: create contentStore: blank space is shown & it shows 3

Farhan-emgage:
FE - Create ContentStore: Blank space is shown & it shows 3 contentdef in a row instead of 4 in step-1 Select Content Template

### DIFF
--- a/src/components/Grid/Grid.scss
+++ b/src/components/Grid/Grid.scss
@@ -17,11 +17,15 @@ $border-color: #dbdbdb;
   
   &.block {
     flex-direction: column;
-    min-width: 180px;
+    min-width: 17.3rem;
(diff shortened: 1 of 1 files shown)


src/components/Grid/Grid.scss
```diff
@@ -17,12 +17,16 @@ $border-color: #dbdbdb;
   
   &.block {
     flex-direction: column;
-    min-width: 180px;
+    min-width: 17.3rem;
```
weegrom, line 20: Can you explain why a min-width value is required.

Farhan-emgage: @weegrom as per our discussion i had removed min-width from grid

PR #481, opened 2020-06-16
fix: id:45121 badge overlapped on content base type icon and name

Farhan-emgage:
FE - Create ContentStore: Entity state badge shows large which is overlapped on content base type icon and contentdef name in step-1 Select Content Template


weegrom:
@Farhan-emgage this is approved, however I did not merge as it looks like this and https://github.com/emgage/emgage-frontend/pull/1074 are related.

PR #504, opened 2020-07-25
Bugfix/44584: dropDown result cutting issue while adding new member o…

CaptainWild:
This PR resolved [44584](https://atera.tpondemand.com/RestUI/Board.aspx#page=board/4941726065867675222&appConfig=eyJhY2lkIjoiOTEwQzZFODEyODVCQ0Y4MTg5RUIyMEJEQTVDMEM2OUYifQ==&boardPopup=bug/44548/silent).

### DIFF
--- a/src/components/Picker/Picker.scss
+++ b/src/components/Picker/Picker.scss
@@ -37,6 +37,7 @@ $stacking-order: (
 }
 
 .container {
+  width: 100%;
   display: inline-block;
   //overflow: hidden;
   text-overflow: ellipsis;


src/components/Picker/AutoSuggestText.tsx
```diff
@@ -38,26 +38,28 @@ class AutoSuggestText extends React.PureComponent<Props, {}> {
...
-          renderSuggestion={this.props.autoSuggestMethods ? this.props.autoSuggestMethods.renderSuggestion : null}
-          highlightFirstSuggestion={true}
-          inputProps={this.props.stateProps ? this.props.stateProps.inputProps : null}
-          ref={this.props.autoSuggestMethods ? this.props.autoSuggestMethods.storeInputReference : this.refHolder}
-          theme={{
-            ...theme,
-            container: theme.container,
-            suggestion: theme.suggestionItem,
-            suggestionsList: theme.suggestionsList,
+        <div style={{ width: '100%' }}>
```
renanpallin, line 41: That's not a good practice, it causes a memory leak within renderes.
One way to accomplish your style is to create a class in pure css and style with "className" property to your div.
You could try remove this div and apply your style to  `theme.suggestionsContainer` as we see in line 43, this way we prevent an additional div.
If not work with suggestionsContainer, just create the css class I said =)

renanpallin:
Be really careful with push force, my friend hahahah

PR #513, opened 2020-08-10
Bugfix/45621 fe message channel time ui shown vertical instead of horizontal

sagar-vaghela:
Added New css file as emage-frontend not include requires css.


emgageadmin:
https://atera.tpondemand.com/entity/45621

PR #522, opened 2020-08-18
fix: id:45995 label ellipses issue in mobile devices

Farhan-emgage:
Apply ellipses to label in mobile devices


emgageadmin:
https://atera.tpondemand.com/entity/45995

PR #523, opened 2020-08-19
Bugfix/46129 fe primitive widgets editable issue

sagar-vaghela:
Hide/remove the Spin button while disabled and read mode for input number


emgageadmin:

CaptainSadaf (approved):

PR #526, opened 2020-08-20
Bugfix/46129 fe primitive widgets editable issue

sagar-vaghela:


emgageadmin:

PR #533, opened 2020-09-09
fix: id:46256 MessageChennel subscriber dropdown overlaping

Farhan-emgage:
MessageChennel subscriber dropdown overlaping with sibling input and select components

### DIFF
--- a/themes/Delicious/Labelled.scss
+++ b/themes/Delicious/Labelled.scss
@@ -1,7 +1,8 @@
 $label-horizontal-spacing: spacing(tighter) * 1.5;
 
 $stacking-order: (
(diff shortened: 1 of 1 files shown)


emgageadmin:
https://atera.tpondemand.com/entity/46256

weegrom (changes requested):
@Farhan-emgage you need to used the staking as defined in https://ateraplat.visualstudio.com/Emgage%20Platform/_wiki/wikis/Emgage%20Platform.wiki/278/Stacking . If label is set to 19 it will likely appears behind the input, which is no good.

weegrom (approved):

PR #538, opened 2020-09-22
feat: 46739 added ID properties for automation script

romesh87:
https://atera.tpondemand.com/entity/46739-fe-need-to-add-id-property


emgageadmin:
https://atera.tpondemand.com/entity/46739

CaptainWild (approved):

PR #543, opened 2020-09-30
fix: id:46394 Icon component accessibility

Farhan-emgage:
Icon accesibility with accesibilityLabel. included role and tabIndex for achieve focus


emgageadmin:
https://atera.tpondemand.com/entity/46394

src/components/Drawer/Drawer.scss
```diff
@@ -280,7 +280,7 @@ body.overlay {
 }
 
 body.container {
-  overflow: auto !important;
+  overflow: visible !important;
```
weegrom, line 283: @Farhan-emgage has this been tested to confirm it does not create an issue with wide content displaying outside the drawer?

weegrom (changes requested):
@Farhan-emgage please respond to my 1 comment, and also resolve merge conflict. Thanks

src/components/Drawer/Drawer.scss
Farhan-emgage: @weegrom yes i tested it. when the drawer is open at that time bodu.container class is apply on body and when drawer is close then this class is removed

weegrom:
@Farhan-emgage please resolve the merge conflict here so we can get this PR merged.

PR #546, opened 2020-10-06
feat: 46157 add create event icon on hover

csumudu:


emgageadmin:
https://atera.tpondemand.com/entity/46157

csumudu:
remove unwanted files

PR #548, opened 2020-10-07
Feature 46708: made type as required field for TextField

emgage-smit:


emgageadmin:
https://atera.tpondemand.com/entity/46708

src/components/TextField/TextField.tsx
```diff
@@ -122,6 +122,10 @@ const getUniqueID = createUniqueIDFactory('TextField');
 class TextField extends React.PureComponent<Props, State> {
   state: State = { height: null, value: '' };
 
+  static defaultProps: any = {
+    type: 'text'
```
weegrom, line 126: @emgage-smit do we want to apply this by default? I thought we decided we wanted to force the dev to select a type value, in which case we would not want a default.

weegrom (approved):

PR #551, opened 2020-10-15
fix: id:42679 apply calendar to match with default theme

Farhan-emgage:
apply  calendar to match with default theme


emgageadmin:
https://atera.tpondemand.com/entity/42679

weegrom (approved):

PR #552, opened 2020-10-15
feature 46708: type="text" added in TextField and ValidatedTextField

emgage-smit:


emgageadmin:
https://atera.tpondemand.com/entity/46708

weegrom (approved):

PR #556, opened 2020-10-20
fix: id:46940 button component accessibility

Farhan-emgage:
button component accessibility

### DIFF
--- a/src/components/Tab/Tab.tsx
+++ b/src/components/Tab/Tab.tsx
@@ -14,20 +14,25 @@ export interface Props {
   activeTabId?: string;
   // Callback function to be called when tab is clicked/selected
   onClick?(): void;
+  onFocus?(): void;
   // User can Set style for Tab component
   componentStyle?: React.CSSProperties;
   // Set theme for tab
   theme?: any;
+  tabIndex?:number;
+  Role?:string;
+  // Descriptive text to be read to screenreaders.
+  accessibilityLabel?: string;
 }
 
 // return a Single tab component to be render into tabpanel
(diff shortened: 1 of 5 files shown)


emgageadmin:
https://atera.tpondemand.com/entity/46940

src/components/Tab/Tab.tsx
```diff
@@ -14,20 +14,25 @@ export interface Props {
   activeTabId?: string;
   // Callback function to be called when tab is clicked/selected
   onClick?(): void;
+  onFocus?(): void;
```
weegrom, line 17: @Farhan-emgage what's the purpose of this prop?

Farhan-emgage: @weegrom this is for tabs. actually this tabs are in <div></div>. and the content is opening on click event. but for accessibility by keyboard it should be open on focus also. because of this reason i added onFocus() prop

weegrom: @Farhan-emgage can you please investigate if there is a way to use a single event for both mouse and keyboard. It seems odd to required both onClick and onFocus. Also, wouldn't OnFocus simply select the tab, and then hitting "Enter" open the tab?

Farhan-emgage: @weegrom Yes, this is possible. but for that we have to place this tab in either in anchor tag or it should be inside of button.

weegrom: @Farhan-emgage what was the final decision here. I'd like to get this PR closed.

PR #562, opened 2020-11-03
fix: id:47493 recipient filter - ui issue when suggestion list is more

Farhan-emgage:
FE - MessageDef - Recipient filter - UI issue when suggestion list is more


emgageadmin:
https://atera.tpondemand.com/entity/47493

HirenTariwala (approved):

PR #566, opened 2020-11-20
fix: id:47815 calendar component accesibility

Farhan-emgage:
calendar component accessibility

### DIFF
--- a/src/components/Calendar/Calendar.css
+++ b/src/components/Calendar/Calendar.css
@@ -277,7 +277,8 @@ button.rbc-input::-moz-focus-inner {
       -webkit-transition: all .3s ease-in-out;
     }
 
-    .rbc-date-cell.rbc-now > a:focus, .rbc-date-cell.rbc-now.rbc-current > a:focus {
+    .rbc-date-cell.rbc-now > a:focus,
+    .rbc-date-cell.rbc-now.rbc-current > a:focus {
       color:#fff;
     }
 
--- a/src/components/Calendar/Calendar.css
+++ b/src/components/Calendar/Calendar.css
@@ -276,6 +276,11 @@ button.rbc-input::-moz-focus-inner {
       -moz-transition: all .3s ease-in-out;
       -webkit-transition: all .3s ease-in-out;
     }
+
+    .rbc-date-cell.rbc-now > a:focus, .rbc-date-cell.rbc-now.rbc-current > a:focus {
+      color:#fff;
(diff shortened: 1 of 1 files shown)


emgageadmin:
https://atera.tpondemand.com/entity/47815

src/components/Calendar/Calendar.css
```diff
@@ -276,6 +276,11 @@ button.rbc-input::-moz-focus-inner {
       -moz-transition: all .3s ease-in-out;
       -webkit-transition: all .3s ease-in-out;
     }
+
+    .rbc-date-cell.rbc-now > a:focus, .rbc-date-cell.rbc-now.rbc-current > a:focus {
```
weegrom, line 280: Add each class declaration on it's own line and remove nesting indents.

weegrom (changes requested):
@Farhan-emgage please see my comment. Take a look at https://github.com/emgage/engage-ui/pull/567 first as it provides greater details as to why these changes are being requested.

src/components/Calendar/Calendar.css
Farhan-emgage: i removed nesting declaration

weegrom: @Farhan-emgage I see that the declaration are now on their own lines, however the code is still nested. Can you please remove the nesting. Thanks

Farhan-emgage: as your comment i understand that you are asking about i declare css property like ".rbc-date-cell.rbc-now > a:focus". and you want to use it like 
.rbc-date-cell{
     .rbc-now{
          a{
            &:focus{
             }
          }
     }
}

weegrom: @Farhan-emgage by nesting I mean:

.classname {
  styles...
}
    .classname {
      styles...
    }
        .classname {
          styles...
        }

Which is very confusing, as it makes it look like the styles are actually nested as is done is scss, like so:

.classname {
  styles...

    .classname {
      styles...

        .classname {
          styles...
        }
    }
}

If they are not actually nested styles, then they should not appear nested, but should have a flat structure like so:

.classname {
  styles...
}
.classname {
  styles...
}
.classname {
  styles...
}

weegrom: @Farhan-emgage can you please take care of the above request so we can get this merged.

weegrom (approved):

PR #569, opened 2020-11-26
fix: id:47681 top bar shrink

Farhan-emgage:
Shrink top Bar


emgageadmin:
https://atera.tpondemand.com/entity/47681

weegrom (approved):

PR #570, opened 2020-11-27
fix: id:47952 Drawer component accessibility

Farhan-emgage:
Drawer component accessibility


emgageadmin:
https://atera.tpondemand.com/entity/47952

PR #572, opened 2020-11-30
fix: id:47992 bugfix for comobobox changes

sagar-vaghela:


emgageadmin:
https://atera.tpondemand.com/entity/47992

ankitsaraogi (approved):

PR #577, opened 2020-12-04
fix: id:48025 - updated logic for hideOnSinglePage

emgage-smit:


emgageadmin:
https://atera.tpondemand.com/entity/48025

HirenTariwala (approved):

PR #582, opened 2020-12-08
fix: id:48167 Tab component HTML structure change

Farhan-emgage:
Tab component HTML structure change

### DIFF
--- a/src/components/Tab/Tab.scss
+++ b/src/components/Tab/Tab.scss
@@ -12,7 +12,8 @@
   border: 2px solid transparent;
 }
 .tab:hover {
-  background-color: #d8d8d8;
+  // background-color: #d8d8d8;
+  border-bottom-color: #DFE3E8;
 }
 .tab.active:hover {
   background-color: transparent;
@@ -21,6 +22,13 @@
   margin-right: 8px;
 }
 
+.tab button {
+  color:#616161;
+}
+.tab button:hover,
+.tab button:focus {
+  color: #212121;
+}
 .tabpanel {
   display: flex;
 }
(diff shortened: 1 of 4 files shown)


emgageadmin:
https://atera.tpondemand.com/entity/48167

weegrom (changes requested):
@Farhan-emgage two things:
1. This needs to use functions for specifying colors (ex: theme-color(m-grey, 900); )
1. This should likely this have a stylesheet added to Delicious theme

Farhan-emgage:
@weegrom i had given variable color instead of static color code. exact color is not found so it might some variation

weegrom (changes requested):
@Farhan-emgage the colors you provided will not work in this file, as this is the base component stylesheet and not the style overrides via the Delicious theme. You're limited to the colors avaible in the base styling, as defined in https://github.com/emgage/engage-ui/blob/master/src/styles/foundation/colors.scss . Please fix this, and also add overrides for the Tab component to Delicious. You can view the Delicious styling for the tab in https://www.figma.com/file/Ff3voHzTudQyla2AAfwmDt/emgage-ui-Components?node-id=915%3A7720

Farhan-emgage:
I added color property in color.scss and apply the variable in Tab.scss

weegrom (changes requested):
@Farhan-emgage  you now have duplicate color vars, as (sky, base) and (grey, lighter) are exactly the same. Please remove the colors you added and instead use:
sky, base (#dfe3e8)
ink, lighter (#637381)
ink, base (#212b36)

You should be able to figure out which color is to be assigned where.

Farhan-emgage:
changed color variable

weegrom (approved):

PR #588, opened 2020-12-14
fix: id:47081 close icon overlapping with scrollbar

CaptainSadaf:


emgageadmin:
https://atera.tpondemand.com/entity/47081

PR #591, opened 2020-12-14
fix: id:47830 remove class from body when drawer is close

Farhan-emgage:
remove class from body when drawer is close

### DIFF
--- a/src/components/Drawer/Drawer.tsx
+++ b/src/components/Drawer/Drawer.tsx
@@ -200,6 +200,8 @@ class Drawer extends React.PureComponent<Props, never> {
 
     if (active && (mode === 'push' || mode === 'reveal')) {
       this.setBodyStyle();
+    } else {
+      document.body.className = '';
     }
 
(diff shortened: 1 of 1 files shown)


emgageadmin:
https://atera.tpondemand.com/entity/47832

src/components/Drawer/Drawer.tsx
```diff
@@ -200,6 +200,8 @@ class Drawer extends React.PureComponent<Props, never> {
 
     if (active && (mode === 'push' || mode === 'reveal')) {
       this.setBodyStyle();
+    } else {
+      document.body.className = '';
```
weegrom, line 204: @Farhan-emgage you can not do this. This is the exact issue you were encountering in messaging, where another component was removing the classes from the body element required by the drawer. It's not appropriate to remove all class names, and there may be another component also reliant upon a class on the body element.

weegrom (changes requested):

PR #600, opened 2020-12-23
fix: id:47832 drawer scroller style change

Farhan-emgage:
drawer scroller styling change


emgageadmin:
https://atera.tpondemand.com/entity/47832

weegrom (approved):
@Farhan-emgage I'm approving this so QA can test, however it looks like you considered only ltr languages, and not rtl. Please test for both ltr and rtl and make any needed adjustments in a separate PR.

PR #603, opened 2020-12-28
fix: id:48337 icon adding minusMinor

CaptainSadaf:


emgageadmin:
https://atera.tpondemand.com/entity/48337

PR #606, opened 2020-12-28
Base color use for switch button

Farhan-emgage:
Base color use for switch button


emgageadmin:
https://atera.tpondemand.com/entity/46978

emgage-smit (approved):

PR #607, opened 2020-12-28
fix: id:48202 item per page alignment issue resolve in SVSU

Farhan-emgage:
item per page alignment issue resolve in SVSU


emgageadmin:
https://atera.tpondemand.com/entity/48202

ankitsaraogi (approved):

PR #613, opened 2021-01-05
Fix: id:48337 add minusMinor icon

CaptainSadaf:


emgageadmin:
https://atera.tpondemand.com/entity/48337

emgage-smit (approved):

PR #626, opened 2021-01-22
fix: id:50162 pagination next/prev button alignment

Farhan-emgage:
pagination next/prev button alignment


emgageadmin:
https://atera.tpondemand.com/entity/50162

weegrom (approved):
@Farhan-emgage I am approving this as a quick fix, however this needs to be set via the button components `ComponentSize` prop and not via custom styles. https://github.com/emgage/engage-ui/blob/master/src/components/Button/Button.tsx#L20

Please make this change. Thanks

PR #627, opened 2021-01-26
id:49877 add minusMinor icon alias

DhruvinPatel78:


emgageadmin:
https://atera.tpondemand.com/entity/49877

src/components/Icon/Icon.tsx
```diff
@@ -201,6 +201,7 @@ export const BUNDLED_ICONS = {
   delete: deleteIcon,
   export: exportIcon,
   import: importIcon,
+  minusIcon: minusMinor,
```
CaptainSadaf, line 204: @DhruvinPatel78 Does icon exist with name minusIcon? I think minusMinor icon exists and an export alias name can be created as minusMinor: minusIcon.

CaptainSadaf (approved):

PR #631, opened 2021-01-27
squad1/bug/49877 added minusMinor in types

DhruvinPatel78:


emgageadmin:
https://atera.tpondemand.com/entity/49877

PR #642, opened 2021-02-01
fix: id: 50290 selection box issue

CaptainSadaf:


emgageadmin:
https://atera.tpondemand.com/entity/50290

PR #647, opened 2021-02-02
fix: id:50353 Alignment issue due to blank list in pagination

Farhan-emgage:
Alignment issue due to blank list in pagination


emgageadmin:
https://atera.tpondemand.com/entity/50353

weegrom (approved):

PR #648, opened 2021-02-02
fix:id:50282 added value prop

DhruvinPatel78:


emgageadmin:
https://atera.tpondemand.com/entity/50282

src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -87,22 +89,22 @@ class DateTimePicker extends React.PureComponent<Props, State>{
...
-    const { dateTime, open } = this.state;
+    const { dateFormat, label, theme, timePicker, placeholder, componentStyle, value } = this.props;
+    const { open } = this.state;
     return (
         <div>
             <TextField
                 type="text"
                 label={label}
-                value={dateTime ? dateTime.format(this.timeFormat) : null}
+                value={value ? moment(value).format(this.timeFormat) : ''}
```
CaptainSadaf, line 99: @DhruvinPatel78  When value prop will not get passed to component then datePicker value remains null on every selection.
if the value prop is undefined in props then keep using dateTime internal state of component otherwise handle it with value prop

CaptainSadaf: You can test it engage-ui test by not passing value in prop.

CaptainSadaf (changes requested):

src/components/DateTimePicker/DateTimePicker.tsx
CaptainSadaf: DataPickerComponent is using in other places of application where value prop is not getting passed.

CaptainSadaf (approved):

PR #651, opened 2021-02-03
fix: id:46394 add visually hidden component in table more action button

Farhan-emgage:
add visually hidden component in table more action button


emgageadmin:
https://atera.tpondemand.com/entity/46394

weegrom (approved):

PR #661, opened 2021-02-08
fix: id:50456 Correct the spelling from Tabuler to Tabular

Farhan-emgage:
Correct the spelling from Tabuler to Tabular


emgageadmin:
https://atera.tpondemand.com/entity/50456

PR #664, opened 2021-02-09
fix: match popover width with figma design

Farhan-emgage:
match popover width with figma design


emgageadmin:

sagar-vaghela (approved):

PR #668, opened 2021-02-11
fix: id:50393 Message Channel message/delivery option alignment issue

Farhan-emgage:
Message Channel message/delivery option alignment issue


emgageadmin:
https://atera.tpondemand.com/entity/50393

weegrom (approved):

PR #673, opened 2021-02-15
fix: id:50494 more action button color change on keyboard focus

Farhan-emgage:
more action button color change on keyboard focus


emgageadmin:
https://atera.tpondemand.com/entity/50494

weegrom (approved):

PR #674, opened 2021-02-16
fix id:45184 fix pages heading styles of text widget

alexmochu:
Before Fix
before_fix

After fix
after_fix


emgageadmin:
https://atera.tpondemand.com/entity/45184

emgage-smit (approved):

PR #676, opened 2021-02-17
fix: id:50594 create bodytext component in next branch

Farhan-emgage:
bodytext component in next branch


emgageadmin:
https://atera.tpondemand.com/entity/50594

PR #678, opened 2021-02-17
fix: id:50580 update npm plugin & implemented picker component

sagar-vaghela:


emgageadmin:

weegrom (approved):

PR #688, opened 2021-02-24
fix: id:50733 switch component accessibility

Farhan-emgage:
switch component accessibility


emgageadmin:
https://atera.tpondemand.com/entity/50733

weegrom (approved):

PR #693, opened 2021-03-01
fix: id:50779 SignOut issue

romesh87:


emgageadmin:
https://atera.tpondemand.com/entity/50779

emgage-smit (approved):

PR #700, opened 2021-03-05
fix: id:50798 sign out clickable area

Farhan-emgage:
sign out clickable area

- add full width to dropdown button
- apply display block to signout anchor
- apply css property to full width button


emgageadmin:
https://atera.tpondemand.com/entity/50798

weegrom (approved):

PR #704, opened 2021-03-10
fix: id:51961 reduce the width of side navigation

Farhan-emgage:
reduce the width of side navigation


emgageadmin:
https://atera.tpondemand.com/entity/51961

PR #707, opened 2021-03-11
fix: id:51961 set alignment of accordion heading if no child

Farhan-emgage:
set alignment of accordion heading if no child


emgageadmin:
https://atera.tpondemand.com/entity/51961

sumeet-jadhav (approved):

PR #708, opened 2021-03-11
fix: id:50798 Sign out button clickable area

Farhan-emgage:
Sign out button clickable area add margin and padding to anchor


emgageadmin:
https://atera.tpondemand.com/entity/50798

weegrom (approved):

PR #709, opened 2021-03-11
fix: id:51809 TextField number incr decr stuck issue, linter fixes

romesh87:


emgageadmin:
https://atera.tpondemand.com/entity/51809

emgage-smit (approved):

PR #715, opened 2021-03-16
fix: id:52023 accordian body overlapping with header

Farhan-emgage:
accordian body overlapping with header

### DIFF
--- a/src/components/SideNavigation/SideNavigation.scss
+++ b/src/components/SideNavigation/SideNavigation.scss
@@ -159,6 +159,8 @@ body .childLi {
     .body {
         padding:0px 0px;
         overflow: visible;
+        z-index: auto;
+        position: relative;
         &>div {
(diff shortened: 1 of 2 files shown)


emgageadmin:
https://atera.tpondemand.com/entity/52023

src/components/SideNavigation/SideNavigation.scss
```diff
@@ -167,4 +169,9 @@ body .childLi {
             }
         }
     }
+
+    .bodyCollapsed {
+        z-index: -10;
```
weegrom, line 174: @Farhan-emgage I'd like to revisit this application of z-index.

Farhan-emgage: @weegrom instead of z index applied display property

weegrom (approved):

PR #720, opened 2021-03-18
bug: 52051: picker double name fix

emgage-smit:


emgageadmin:
https://atera.tpondemand.com/entity/52051

Farhan-emgage (approved):

PR #729, opened 2021-03-23
fix: id:51905 issue for the picker

sagar-vaghela:


emgageadmin:
https://atera.tpondemand.com/entity/51905

emgage-smit (approved):

PR #733, opened 2021-03-24
fix: id:52069 More action option styling in app list tree view

Farhan-emgage:
More action option styling in app list tree view


emgageadmin:
https://atera.tpondemand.com/entity/52069

weegrom (approved):

PR #746, opened 2021-03-30
fix: id:52033 Error border to text area

Farhan-emgage:
Error border to text area


emgageadmin:
https://atera.tpondemand.com/entity/52033

ankitsaraogi (approved):

PR #747, opened 2021-03-30
fix: side navigation should re render after props are changed

sumeet-jadhav:


src/components/SideNavigation/SideNavigation.tsx
```diff
@@ -91,11 +99,6 @@ class SideNavigation extends React.PureComponent<Props, State> {
     if (newProps.drawerExpand !== this.props.drawerExpand) {
```
sagar-vaghela, line 99: Put Comments which removed

sagar-vaghela (approved):

PR #748, opened 2021-03-30
fix: id:52323 SideNavigation state not saved, linter fixes

romesh87:


emgageadmin:
https://atera.tpondemand.com/entity/52323

emgage-smit (approved):

PR #755, opened 2021-04-05
fix: id:52520 DropdownItem disabled styles issue

romesh87:


emgageadmin:
https://atera.tpondemand.com/entity/52520

emgage-smit (approved):

PR #759, opened 2021-04-08
fix: id:52463 Dropdown width issue resolve in picker with more button

Farhan-emgage:
Dropdown width issue resolve in picker with more button

### DIFF
--- a/src/components/Picker/Picker.scss
+++ b/src/components/Picker/Picker.scss
@@ -80,6 +80,7 @@ $stacking-order: (
   box-shadow: 0 2px 5px 0 rgba(0, 0, 0, 0.25);
   list-style-type: none;
   z-index: z-index(popover, $stacking-order);
+  min-width: 13rem;
 }
 
 
@@ -126,7 +127,6 @@ $stacking-order: (
 .PopoverButtonWrap {
   min-height: 30px;
   position: absolute;
-  max-width: 42rem;
   height: 100%;
   .suggestionsList {
     width:100%;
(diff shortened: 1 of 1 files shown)


emgageadmin:
https://atera.tpondemand.com/entity/52463

src/components/Picker/Picker.scss
```diff
@@ -126,7 +126,7 @@ $stacking-order: (
 .PopoverButtonWrap {
   min-height: 30px;
   position: absolute;
-  min-width: 42rem;
+  max-width: 42rem;
```
weegrom, line 129: Hi @Farhan-emgage this is not the desired fix, as it will wrap long names in the picker. Can you please look into this more. Thanks

Farhan-emgage: @weegrom actually main issue is previously the popover of picker has no width. but when button appear at that time min-width is applying. because of that the popover is creating issue with width.

weegrom: @Farhan-emgage setting `max-width: 42rem;` is not the appropriate fix. In this case a picker option over 42rem will not display properly or wrap. We set 42rem min width to keep picker options such as [A, B, C, D, F] from displaying in dropdown that's so thin it's difficult to interact with. The dropdown width should be based on it's content, and only limitied by the browser screen width.

Farhan-emgage: ok @weegrom will look after it

weegrom (approved):

PR #760, opened 2021-04-08
fix: id:52620 set hover and focus changes in side navigation

Farhan-emgage:
set hover and focus changes in side navigation


emgageadmin:
https://atera.tpondemand.com/entity/52620

weegrom (approved):

PR #761, opened 2021-04-08
fix: id:52636 text field height issue when text enter

Farhan-emgage:
text field height issue when text enter


emgageadmin:
https://atera.tpondemand.com/entity/52636

weegrom (approved):

PR #763, opened 2021-04-13
fix:id:52590 icon prop added

DhruvinPatel78:


emgageadmin:
https://atera.tpondemand.com/entity/52590

PR #767, opened 2021-04-16
fix: id:52620 focus effect on main heading of sidebar

Farhan-emgage:
focus effect on main heading of sidebar


emgageadmin:
https://atera.tpondemand.com/entity/52620

weegrom (approved):

PR #768, opened 2021-04-19
fix: id:52800 remove after from desktop and add in mobile in table tr

Farhan-emgage:
remove after from desktop and add in mobile in table tr


emgageadmin:
https://atera.tpondemand.com/entity/52800

weegrom (approved):

PR #771, opened 2021-04-22
fix:id:52801 image styling prop added

DhruvinPatel78:


emgageadmin:
https://atera.tpondemand.com/entity/52801

CaptainSadaf (approved):

PR #776, opened 2021-04-27
fix: id:52913 UI issue resolve when subscriber name is long

Farhan-emgage:
UI issue resolve when subscriber name is long


emgageadmin:
https://atera.tpondemand.com/entity/52913

sagar-vaghela (approved):

PR #777, opened 2021-04-27
fix: id:52979 remove font weight from devider styling in list

Farhan-emgage:
remove font weight from devider styling in list


emgageadmin:
https://atera.tpondemand.com/entity/52979

weegrom (approved):

PR #779, opened 2021-04-27
fix: id:52639 Set styling of table and width of modal in small

Farhan-emgage:
Set styling of table and width of modal in small


emgageadmin:
https://atera.tpondemand.com/entity/52639

weegrom (approved):

PR #784, opened 2021-04-29
fix: id:52979 remove font weight to span childs

Farhan-emgage:
remove font weight to span childs


emgageadmin:
https://atera.tpondemand.com/entity/52979

sagar-vaghela (approved):

PR #794, opened 2021-05-11
fix: id:53190 Login button width issue

Farhan-emgage:
Login button width issue


emgageadmin:
https://atera.tpondemand.com/entity/53190

weegrom (approved):

PR #797, opened 2021-05-12
fix: id:53157 changes in width on picker component popover

Farhan-emgage:
Changes in width on picker component popover


emgageadmin:
https://atera.tpondemand.com/entity/53157

sagar-vaghela (approved):

PR #802, opened 2021-05-17
fix: id:53157 dropdown issue resolve in message recipient picker dropdown

Farhan-emgage:
dropdown issue resolve in message recipient picker dropdown


emgageadmin:
https://atera.tpondemand.com/entity/53157

emgage-smit (approved):

PR #810, opened 2021-05-20
fix: id:53325 Text area content overlap with its label text

Farhan-emgage:
Text area content overlap with its label text


emgageadmin:
https://atera.tpondemand.com/entity/53325

weegrom (approved):

PR #814, opened 2021-05-24
fix: id:53463 add button in chip icons and other corrections

Farhan-emgage:
add button in chip icons and other corrections


emgageadmin:
https://atera.tpondemand.com/entity/53463

src/components/Chip/Chip.scss
```diff
@@ -21,6 +21,38 @@ $vertical-padding: ($min-height - line-height(body) - rem(2px)) / 2;
...
+        padding: 0px;
+        min-width: auto;
+        min-height: auto;
+        margin:0px;
+        vertical-align: middle;
+        padding-top: 0.6rem;
+        padding-bottom: 0.5rem;
+        padding-left: 3.5rem;
+        position: relative;
+        z-index: 10;
```
weegrom, line 37: @Farhan-emgage is this z-index needed? I'm concerned it may lead to the chip displaying on top of other items such as dropdowns.

Farhan-emgage: @weegrom actually i placed this z index in respective to image  beside text. let me check is it creating issue with other component.

Farhan-emgage: @weegrom 

@emgage-smit has done the required changes

src/components/Chip/Chip.scss
```diff
@@ -40,13 +71,28 @@ $vertical-padding: ($min-height - line-height(body) - rem(2px)) / 2;
...
+    border-radius: 50%; 
+    display: inline-block;  
+    vertical-align: middle;
 }
 
-.Remove {
-    cursor: pointer;
-    flex-shrink: 0;
-    margin: ($horizontal-padding / 2) (-1 * $horizontal-padding / 2) ($horizontal-padding / 2) ($horizontal-padding / 2);
+body .Remove {
```
weegrom, line 79: @Farhan-emgage why was body added to this selector? Also, color's should be expressed using the color() mixin and margin/padding values typically use variables to allow for consistency and easy adjustment.

src/components/Chip/Chip.scss
```diff
@@ -12,7 +12,7 @@ $vertical-padding: ($min-height - line-height(body) - rem(2px)) / 2;
     height: $height;
     min-width: $height;
     margin: 0;
-    padding: 0 $horizontal-padding;
+    padding: 0 ($horizontal-padding + .3) 0 0;
```
weegrom, line 15: @Farhan-emgage why is the right padding being increased and the left padding removed? I'm pretty sure this will break instances of the chip with image prop, and make the left portion of the newly added chipButtonContent unclickable.

weegrom (commented):
@Farhan-emgage please schedule a meeting to go over your changes, as is see potential issues and divergence from our normal practices.

src/components/Chip/Chip.scss
Farhan-emgage: @weegrom  Actually previously remove icon was in link. but now we took it in button. For making remove button alignment with sibling button we have to add some padding. but in chip component make changes for button style there are two ways either we have to place important or we apply strong relation. so i apply body before remove class.

src/components/Chip/Chip.scss
Farhan-emgage: @weegrom  actually previously image / icon (left side) and content both was separate but now we place it in one button. so for that we had remove the left padding. and for right padding there are border between remove and content. so making space between them we had increase the right padding

PR #817, opened 2021-05-26
fix: id:53486 changes in style when user click beside the choice text

Farhan-emgage:
changes in style when user click beside the choice text


emgageadmin:
https://atera.tpondemand.com/entity/53486

weegrom (approved):

PR #824, opened 2021-05-31
fix: id:53548 BreadCrumb pointer issue

romesh87:


emgageadmin:
https://atera.tpondemand.com/entity/53548

emgage-smit (approved):

PR #829, opened 2021-06-07
Userstory/53654 ux create feedback icons

weegrom:


emgageadmin:
https://atera.tpondemand.com/entity/53654

PR #835, opened 2021-06-10
fix: id:53575 add scale in body to reduce hide content

Farhan-emgage:
add scale in body to reduce hide content


emgageadmin:
https://atera.tpondemand.com/entity/53575

weegrom (approved):

PR #837, opened 2021-06-10
fix: id:53701 checkbox helptext style fix

romesh87:


emgageadmin:
https://atera.tpondemand.com/entity/53701

emgage-smit (approved):

PR #839, opened 2021-06-11
fix: id:53700 Select disabled options added

romesh87:


emgageadmin:
https://atera.tpondemand.com/entity/53700

emgage-smit (approved):

PR #842, opened 2021-06-15
fix: id:52929 icon text alignment in dropdown

Farhan-emgage:
icon text alignment in dropdown


emgageadmin:
https://atera.tpondemand.com/entity/52929

weegrom (approved):

PR #843, opened 2021-06-16
fix: id:53756 remove scale property from scale media query

Farhan-emgage:
remove scale property from scale media query


emgageadmin:
https://atera.tpondemand.com/entity/53756

weegrom (approved):

PR #846, opened 2021-06-24
fix: id:53962 span element added to BodyText component

Farhan-emgage:
span element added to BodyText component


emgageadmin:
https://atera.tpondemand.com/entity/53962

weegrom (approved):

PR #852, opened 2021-06-30
fix: id:53621 popover alignment in login profile click

Farhan-emgage:
popover alignment in login profile click


emgageadmin:
https://atera.tpondemand.com/entity/53621

weegrom (approved):

PR #863, opened 2021-07-08
fix: id:53463 spacing correction in chip component

Farhan-emgage:


emgageadmin:
https://atera.tpondemand.com/entity/53463

emgage-smit (approved):

PR #864, opened 2021-07-08
fix: id:54164 textfield props correction

Farhan-emgage:
textfield props correction


emgageadmin:
https://atera.tpondemand.com/entity/54164

emgage-smit (approved):

PR #872, opened 2021-07-14
fix: id:54277 bugfix for picker component

sagar-vaghela:


emgageadmin:
https://atera.tpondemand.com/entity/54277

emgage-smit (approved):

PR #873, opened 2021-07-15
fix: id:54307 description width issue when max character apply in table

Farhan-emgage:
description width issue when max character apply in table


emgageadmin:
https://atera.tpondemand.com/entity/54307

emgage-smit (approved):

PR #874, opened 2021-07-19
fix: id:54199 MessageDef read UI correections

Farhan-emgage:


emgageadmin:
https://atera.tpondemand.com/entity/54199

sagar-vaghela (approved):

PR #875, opened 2021-07-19
fix: id:54323 UI correction in contentDef selection

Farhan-emgage:
UI correction in contentDef selection


emgageadmin:
https://atera.tpondemand.com/entity/54323

sagar-vaghela (approved):

PR #881, opened 2021-07-23
fix: id:53705 hover styling correction in chip component

Farhan-emgage:
hover styling correction in chip component


emgageadmin:
https://atera.tpondemand.com/entity/53705

emgage-smit (approved):

PR #886, opened 2021-07-27
fix: id:54386,54352 prefix in text field and select in read only

Farhan-emgage:
prefix in text field and select in read only


emgageadmin:
https://atera.tpondemand.com/entity/54352

emgage-smit (approved):

PR #888, opened 2021-07-29
fix: id:54448 More action alignment with long description in table

Farhan-emgage:
More action alignment with long description in table


emgageadmin:
https://atera.tpondemand.com/entity/54448

weegrom (approved):

PR #890, opened 2021-07-30
fix: id:54474 Number field button click select text

Farhan-emgage:
Number field button click select text


emgageadmin:
https://atera.tpondemand.com/entity/54474

weegrom (approved):

PR #902, opened 2021-08-12
fix: id:54607 add text in next previous button in pagination

Farhan-emgage:
add text in next previous button in pagination

### DIFF
--- a/src/components/NoData/NoData.tsx
+++ b/src/components/NoData/NoData.tsx
@@ -51,7 +51,7 @@ const NoData = ({
             }
 
         </div>
-    );
+  );
 };
 
 export default themr(NODATA, baseTheme)(NoData) as ThemedComponentClass<Props, {}>;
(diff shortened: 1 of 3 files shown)


emgageadmin:
https://atera.tpondemand.com/entity/54607

weegrom (changes requested):
Hey @Farhan-emgage - Couple changes:

1. Simplify the text to "Previous" and "Next"
2. On the ul element wrapping the pagination include an aria label or visually hidden text "Pagination". This way a user using a screen reading will know what the list of items is/does/

Thanks

Farhan-emgage:
@weegrom  the changes is done

weegrom (commented):
@Farhan-emgage here also, why are we not using the button components accessibilityLabel prop?

Farhan-emgage:
@weegrom actually accesibilityLabel props is use for screen read. but here we are placing visually hidden component because icon is not considered as text. so screen consider it as a blank button.

image

weegrom:
@Farhan-emgage shouldn't adding the "aria-label" (accessibilityLabel prop) be sufficient? If not, we need to make an adjustment to the button component to make it accessible. We should not require the dev to add a visually hidden component if they've already added the accessibilityLabel prop.

PR #912, opened 2021-08-18
fix: id:54675 more action popover scroller remove

Farhan-emgage:
more action popover scroller remove


emgageadmin:
https://atera.tpondemand.com/entity/54675

DhruvinPatel78 (approved):

PR #917, opened 2021-08-25
fix: id:54606 remove visuallyhidden and  add title in login component

Farhan-emgage:
remove visuallyhidden and  add title in login component


emgageadmin:
https://atera.tpondemand.com/entity/54606

emgage-smit (approved):

PR #918, opened 2021-08-25
fix: id:54617 content item  button popover alignment in small screen

Farhan-emgage:
content item  button popover alignment in small screen


emgageadmin:
https://atera.tpondemand.com/entity/54617

emgage-smit (approved):

PR #919, opened 2021-08-25
fix: id:54614 Add hidden text in more action colum header in table

Farhan-emgage:
Add hidden text in more action colum header in table


emgageadmin:
https://atera.tpondemand.com/entity/54614

weegrom (approved):

PR #925, opened 2021-08-31
fix: id:54700 color mismatch solution for disable and read-only text field

Farhan-emgage:
color mismatch solution for disable and read-only text field


emgageadmin:
https://atera.tpondemand.com/entity/54700

weegrom (changes requested):
These are not the correct colors when using the Delicious theme. This needs to match the color as provided in the mockups and work item comments.

PR #934, opened 2021-09-01
fix: id:54480 add archive and archiveSolid icons

weegrom:


emgageadmin:
https://atera.tpondemand.com/entity/54480

PR #936, opened 2021-09-02
EntityState component: css update

emgage-smit:


emgageadmin:
https://atera.tpondemand.com/entity/54480

PR #945, opened 2021-09-14
fix: id:54974 disable textfield display backdrop

Farhan-emgage:
disable textfield display backdrop


weegrom (commented):
@Farhan-emgage this is not a change we want. When a field is disabled it is to show the backdrop with a light gray background, as defined farther down this stylesheet by the .disabled > .backdrop declaration. You should not be changing the disabled backdrop to transparent. Please delete this PR.

emgageadmin:
https://atera.tpondemand.com/entity/54974

PR #948, opened 2021-09-16
fix: id:54984 remove static width from combobox popover

Farhan-emgage:
remove static width from combobox popover


emgageadmin:
https://atera.tpondemand.com/entity/54984

sagar-vaghela (approved):

PR #953, opened 2021-09-21
fix: id:55031 combobox accessibility

Farhan-emgage:
combobox accessibility


emgageadmin:
https://atera.tpondemand.com/entity/55031

PR #956, opened 2021-09-28
fix: id:55102 combobox item selection scroll page issue

Farhan-emgage:
combobox item selection scroll page issue


emgageadmin:
https://atera.tpondemand.com/entity/55102

PR #959, opened 2021-10-14
fix: id:55116 revert code for auto complete in combo box

Farhan-emgage:
revert code for auto complete in combo box


emgageadmin:
https://atera.tpondemand.com/entity/55116

PR #960, opened 2021-10-14
fix: id:55317 link color change in delicious theme

Farhan-emgage:
link color change in delicious theme


emgageadmin:
https://atera.tpondemand.com/entity/55317

PR #963, opened 2021-10-18
fix: id:54788 corrections in switch component

Farhan-emgage:
corrections in switch component


emgageadmin:
https://atera.tpondemand.com/entity/54788

PR #965, opened 2021-10-18
fix: id:55033 correction in switch circle alignment

Farhan-emgage:
correction in switch circle alignment


emgageadmin:
https://atera.tpondemand.com/entity/55033

PR #979, opened 2021-11-08
fix: id:55538 long description name warped in table

Farhan-emgage:
long description name warped in table

### DIFF
--- a/src/components/Table/Table.scss
+++ b/src/components/Table/Table.scss
@@ -33,7 +33,7 @@ $border-color: color(sky, base);
     .subText {
       margin-top: 2px !important;
       margin-bottom: 0px !important;
-      word-break: break-all;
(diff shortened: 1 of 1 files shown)


emgageadmin:
https://atera.tpondemand.com/entity/55538

src/components/Table/Table.scss
```diff
@@ -33,6 +33,7 @@ $border-color: color(sky, base);
     .subText {
       margin-top: 2px !important;
       margin-bottom: 0px !important;
+      word-break: break-all;
```
weegrom, line 36: @Farhan-emgage I don't think this will work, as it will even break small words instead of wrapping them. Try overflow-wrap: break-word;

weegrom (changes requested):

weegrom (approved):

PR #982, opened 2021-11-11
fix: id:55584 text color change in disable switch box

Farhan-emgage:
text color change in disable switch box


emgageadmin:
https://atera.tpondemand.com/entity/55584

emgage-smit (approved):

PR #994, opened 2021-11-24
fix: id:55501 change element type in banner sub text

Farhan-emgage:
change element type in banner sub text


emgageadmin:
https://atera.tpondemand.com/entity/55501

sagar-vaghela (approved):

PR #995, opened 2021-11-25
fix: id:55109 change option count color in facet component

Farhan-emgage:
change option count color in facet component


emgageadmin:
https://atera.tpondemand.com/entity/55109

emgage-smit (approved):

PR #998, opened 2021-11-29
fix: id:55604 selection display on long press on up/down arrow combobox

Farhan-emgage:
selection display on long press on up/down arrow combobox


emgageadmin:
https://atera.tpondemand.com/entity/55604

sagar-vaghela (approved):

PR #1003, opened 2021-12-06
fix: id:55692 hover/focus styling change in segment component

Farhan-emgage:
hover/focus styling change in segment component


emgageadmin:
https://atera.tpondemand.com/entity/55692

emgage-smit (approved):

PR #1008, opened 2021-12-13
fix: id:55842 create vertical view of tab

Farhan-emgage:
create vertical view of tab


emgageadmin:

emgage-smit (approved):

PR #1010, opened 2021-12-13
fix: id:55842 border corner corrections

Farhan-emgage:
border corner corrections


emgageadmin:

emgage-smit (approved):

PR #1022, opened 2021-12-16
fix: id:55841 add new icon

Farhan-emgage:
add new icons


emgageadmin:
https://atera.tpondemand.com/entity/55841

DhruvinPatel78 (approved):

PR #1024, opened 2021-12-17
fix: id:55841 remove padding from process ul

Farhan-emgage:
remove padding from process ul


emgageadmin:
https://atera.tpondemand.com/entity/55841

DhruvinPatel78 (approved):

PR #1025, opened 2021-12-23
fix: id:55842 add tooltip to tab component

Farhan-emgage:
add tooltip to tab component


emgageadmin:

weegrom:
@Farhan-emgage - a "showTooltip" bool type prop should be added to provide a choice between showing and not showing the button tooltips.

weegrom (changes requested):
@Farhan-emgage - a "showTooltip" bool type prop should be added the Tab component to provide an choice between showing and not showing the button tooltips.

PR #1030, opened 2022-01-13
fix: id:56037 Unit convert PX to REM

Farhan-emgage:
Unit convert PX to REM


Farhan-emgage:
no actually there was some confusion which is clear on yesterdays discussion so now you can commit. but we have to check this changes effect on front end and backend

weegrom:
@Farhan-emgage thanks for the confirmation. This PR has been merged.

